PR #3, opened 2020-08-22
feat: set up first components Loader and Header (tbd)

pyphilia:


pyphilia:
@juancarlosfarah @hagoptaminian This PR is now working. After `yarn` and `yarn start` in `example` (and maybe `yarn` in root) you should see a Header.

pyphilia:
Commit https://github.com/graasp/graasp-ui/pull/3/commits/fac2dd41c6ec48ee0f592139dea1a635988cc5bd handles test dependencies. 

`react` and other modules cannot be in `devDependencies` otherwise it will create hooks problems (react problems) in the projects where it is installed in. On the other hand, jest needs these modules to be able to run tests, so the current solution is to map these modules on `./example` dependencies. Thus it requires to run `yarn` in `example` before running tests, which is handled by the newly added `pretest` script.

juancarlosfarah (approved):
Looks good!

PR #5, opened 2020-09-24
chore: add husky

pyphilia:


.gitignore
```diff
@@ -21,3 +21,7 @@ yarn-error.log*
 
 # development environments
 .idea/
+
+# react-app-env
+src/react-app-env.d.ts
```
juancarlosfarah, line 26: This is new, right?

juancarlosfarah (approved):
Looks good to me! Just one question: what is the `react-app-env.d.ts` file?

pyphilia:
@juancarlosfarah From this issue https://github.com/facebook/create-react-app/issues/6560 It is an automatically generated file because of typescript. It seems necessary yet it contains nothing. Many included it in gitignore, and since it generates automatically it should not be a problem. But we could also include it in the project, let me know what you think.

juancarlosfarah:
@pyphilia, no problem. Sounds like it should be ignored!

PR #9, opened 2021-05-25
ci: add workflow to publish to npm

### LINKED ISSUE #8: Add github actions to publish automatically to npm
https://github.com/marketplace/actions/npm-publish

pyphilia:
close #8

### DIFF
--- a/.github/workflows/publishNpm.yml
+++ b/.github/workflows/publishNpm.yml
@@ -17,6 +17,6 @@ jobs:
       - run: yarn test
 
       - name: publish to npm
-      - uses: JS-DevTools/npm-publish@v1
+        uses: JS-DevTools/npm-publish@v1
         with:
           token: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publishNpm.yml
+++ b/.github/workflows/publishNpm.yml
@@ -1,8 +1,9 @@
 on:
   push:
-    # Sequence of patterns matched against refs/tags
+    # sequence of patterns matched against refs/tags
     tags:
-      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+      # push events to matching v*, i.e. v1.0, v20.15.10
+      - 'v*' 
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,7 +36,8 @@
   "peerDependencies": {
     "@material-ui/core": "4.11.0",
     "@material-ui/icons": "4.9.1",
-    "react": "^16.0.0"
+    "react": "^16.13.1",
```
juancarlosfarah, line 39: Should this be `^`? I forget what's needed for peerDeps.

.github/workflows/publishNpm.yml
```diff
@@ -0,0 +1,21 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
```
juancarlosfarah, line 5: Keep comments above, if possible and all small caps.

.github/workflows/publishNpm.yml
```diff
@@ -0,0 +1,21 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
```
juancarlosfarah, line 3: Minor thing: try to keep all comments lower case.

.github/workflows/publishNpm.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+      - run: yarn install
+      - run: yarn test
+
+      - name: Publish to NPM
```
juancarlosfarah, line 18: Minor thing. Let's all the names lowercase, if possible.

juancarlosfarah (changes requested):
Thanks for this! Just some very very minor comments and a question.

package.json
pyphilia: I think that is okay to do this way. I checked material-ui and there's `^` in their peerDependencies https://github.com/mui-org/material-ui-pickers/blob/e875729b3492a4aedebec9f14589accc83475241/lib/package.json

pyphilia:
Before merging, I will need a NPM_TOKEN. @juancarlosfarah  do we have a graasp account or should we use a token generated by one of us?

juancarlosfarah:
@pyphilia we have an account. I believe you should be a part of it.

PR #11, opened 2021-06-04
10/dynamic tree view

### LINKED ISSUE #10: Add Dynamic Tree View

pyphilia:
This PR adds a `DynamicTreeView` which is a navigation tree. This can have checkboxes (or not) and fetches the children dynamically given react query hooks.

The following is an example integrated within graasp-compose.

Screenshot 2021-06-04 at 10 46 22

close #10


src/DynamicTreeView/DynamicTreeView.js
```diff
@@ -0,0 +1,104 @@
...
+  );
+};
+
+DynamicTreeView.defaultProps = {
+  showItemFilter: () => true,
+  showCheckbox: true,
+  initialExpendedItems: [],
+  shouldFetchChildrenForItem: () => true,
+  isTreeItemDisabled: () => false,
+  buildTreeItemClass: () => '',
```
juancarlosfarah, line 101: What's the difference b/w returning the values or just defining them? Seems to me like it's inconsistent. Also, do we need `propTypes`?

juancarlosfarah (approved):
Looks good! Just one question. 👍

src/DynamicTreeView/DynamicTreeView.js
pyphilia: These functions are mostly used by tree item and depends on the node's id, type, etc. We could accept `boolean` or `function`

`PropTypes` are not more necessary, it is defined in `DynamicTree.d.ts`

PR #13, opened 2021-06-09
12/item components

### LINKED ISSUE #12: Add Item components

pyphilia:
close #12

### DIFF
--- a/src/items/AppItem.js
+++ b/src/items/AppItem.js
@@ -1,7 +1,6 @@
-import React, { useRef, useEffect } from 'react';
-import PropTypes from 'prop-types';
-import { Map } from 'immutable';
+import React, { useRef, useEffect, useState } from 'react';
 import { getAppExtra } from '../utils/itemExtra';
+import Loader from '../Loader';
 
 export const GET_AUTH_TOKEN = 'GET_AUTH_TOKEN';
 export const GET_AUTH_TOKEN_SUCCEEDED = 'GET_AUTH_TOKEN_SUCCEEDED';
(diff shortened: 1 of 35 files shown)


src/items/AppItem.js
```diff
@@ -0,0 +1,121 @@
...
+      {iframeIsLoading && <Loader />}
+      <iframe
+        id={id}
+        title={item?.get('name')}
+        onLoad={() => setIframeIsLoading(false)}
+        ref={iframeRef}
+        width='100%'
+        height={300}
+        src={url}
+        frameBorder={0}
```
juancarlosfarah, line 115: We should probably factor out these constants.

src/items/DownloadButtonFileItem.js
```diff
@@ -0,0 +1,38 @@
...
+      target='_blank'
+      rel='noreferrer'
+      download={name}
+    >
+      <Button
+        color='primary'
+        variant='contained'
+        size='large'
+        startIcon={<GetAppIcon />}
+      >{`Download ${name}`}</Button>
```
juancarlosfarah, line 29: Do we not have translation in UI? Or are we leaving that to each implementation?

src/items/FileItem.js
```diff
@@ -0,0 +1,75 @@
...
+      name={originalFileName}
+      url={url}
+      defaultText={defaultText}
+    />
+  );
+};
+
+FileItem.defaultProps = {
+  defaultItem: null,
+  defaultText: 'No preview available for this file',
```
juancarlosfarah, line 72: Translation? (Same as previous comment).

src/items/S3FileItem.js
```diff
@@ -0,0 +1,77 @@
...
+      name={originalFileName}
+      url={url}
+      defaultText={defaultText}
+    />
+  );
+};
+
+S3FileItem.defaultProps = {
+  defaultItem: null,
+  defaultText: 'No preview available for this file',
```
juancarlosfarah, line 74: Translation? Same as previous comment.

juancarlosfarah (changes requested):
Everything looks good! Just a couple of minor comments and questions. 🏑

src/items/DownloadButtonFileItem.js
pyphilia: We don't have translation, I think it is better to let the user provide a string, otherwise it defaults to english/set text. ideally it would be nice to provide all languages and depends it on the implementation, but it would require to export an instance of `i18n` and I'm not sure it is easy to extends with multiple language files... It might be interesting to investigate at some point.

PR #15, opened 2021-06-16
feat: add captions to items

### LINKED ISSUE #14: Add Caption to Items

pyphilia:
close #14

### DIFF
--- a/src/TextEditor/TextEditor.js
+++ b/src/TextEditor/TextEditor.js
@@ -3,7 +3,6 @@ import { makeStyles } from '@material-ui/core';
 import ReactQuill from 'react-quill';
 import 'react-quill/dist/quill.snow.css';
 import PropTypes from 'prop-types';
-// import Button from '@material-ui/core/Button';
 import { TEXT_EDITOR_MIN_HEIGHT, TEXT_EDITOR_TOOLBAR } from '../constants';
 
 // formula dependencies
@@ -17,7 +16,8 @@ const TextEditor = ({
   value: initialValue,
   edit,
   placeholderText,
-  onSave,
+  onBlur,
+  onChange,
 }) => {
   // keep current content
   const [value, setValue] = useState(initialValue);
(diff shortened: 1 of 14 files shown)


src/TextEditor/TextEditor.js
```diff
@@ -11,34 +11,59 @@ import 'katex/dist/katex.min.css';
...
+      setValue(v);
+    }
+    // eslint-disable-next-line no-unused-expressions
+    onChange?.(v);
+  };
+
   const classes = useStyles();
 
-  const placeholder = !readOnly && placeholderText ? 'Write something' : null;
+  const placeholder = edit && placeholderText ? 'Write something' : null;
```
juancarlosfarah, line 48: Do we need translation here?

src/items/AppItem.js
```diff
@@ -122,6 +131,20 @@ const AppItem = ({ id, item, user, apiHost }) => {
       />
     </React.Fragment>
   );
+
+  if (showCaption) {
```
juancarlosfarah, line 135: Just to confirm, we agreed that the caption always goes below, right? Otherwise we have to add something to switch it from below to above.

src/items/LinkItem.js
```diff
@@ -9,31 +10,63 @@ const useStyles = makeStyles(() => ({
...
+  onSaveCaption,
+  editCaption,
+  showCaption,
+}) => {
   const classes = useStyles();
 
   const id = item.get('id');
   const extra = getEmbeddedLinkExtra(item.get('extra'));
 
+  const Wrapper = withCaption({
```
juancarlosfarah, line 25: Why did you call this `Wrapper` and not `Caption` or `caption`?

juancarlosfarah (changes requested):
👍 

Just a couple question and minor comments.

src/items/AppItem.js
pyphilia: Exactly, but this setting does not exist yet. We decided with André to have this display setting in `item.setting`. So I propose to have this feature in another PR.

PR #18, opened 2021-06-22
fix: refactor AppItem as Component, reset channel, avoid rendering

### LINKED ISSUE #17: Refactor AppItem
Because `AppItem` is rendering multiple times / is mount/unmount in the interface, the handshake protocol with the app get interrupted. This leads for the app to stop asking for the context/token and never has access to the app data.

- Rendering can be avoided with a class-component.
- The parent window should always listen and responds when the context is requested. In that case, the message channel should be reset.

### LINKED ISSUE #19: Show Error when url is missing in File and S3File

pyphilia:
close #17 
close #19

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -27,3 +27,4 @@ export const TEXT_EDITOR_MIN_HEIGHT = 200;
 export const APP_ITEM_HEIGHT = 300;
 export const APP_ITEM_WIDTH = '100%';
 export const APP_ITEM_FRAME_BORDER = 0;
+export const UNEXPECTED_ERROR_MESSAGE = 'An unexpected error occurred';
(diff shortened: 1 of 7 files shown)


src/constants.ts
```diff
@@ -27,3 +27,4 @@ export const TEXT_EDITOR_MIN_HEIGHT = 200;
 export const APP_ITEM_HEIGHT = 300;
 export const APP_ITEM_WIDTH = '100%';
 export const APP_ITEM_FRAME_BORDER = 0;
+export const UNEXPECTED_ERROR_MESSAGE = 'An unexpected error occurred';
```
juancarlosfarah, line 30: Do we need translation here?

src/items/FileItem.js
```diff
@@ -41,7 +43,7 @@ const FileItem = ({
   }, [content]);
 
   if (!url) {
-    return null;
+    return <Alert severity='error'>{errorMessage}</Alert>;
```
juancarlosfarah, line 46: Maybe the translate can go here?

juancarlosfarah (approved):
Looking good! Just a quick question about translate. I forget if it goes here or if we do it at the client level (compose/perform/etc.).

PR #40, opened 2021-07-29
39/yarn2

### LINKED ISSUE #39: Migrate to Yarn2

pyphilia:
close #39


package.json
```diff
@@ -17,7 +17,8 @@
   "scripts": {
     "build": "microbundle-crl --no-compress --format modern,cjs",
     "start": "microbundle-crl watch --no-compress --format modern,cjs",
-    "prepare": "run-s build",
+    "prepack": "run-s build",
+    "prepare":"yarn prepack",
```
juancarlosfarah, line 21: Missing space after `:`.

src/items/AppItem.js
```diff
@@ -16,6 +16,7 @@ export const GET_CONTEXT = 'GET_CONTEXT';
 export const GET_CONTEXT_SUCCEEDED = 'GET_CONTEXT_SUCCEEDED';
 
 const requestApiAccessToken = async ({ id, origin, app, apiHost }) => {
+  // eslint-disable-next-line no-undef
```
juancarlosfarah, line 19: Is this because of `fetch`? I think you should be able to tell ESLint that we're in a browser context and it should see `fetch`.

src/items/AppItem.js
```diff
@@ -119,6 +120,7 @@ class AppItem extends Component {
     if (type === GET_CONTEXT) {
       // create/reset channel and
       // Listen for messages on port1
+      // eslint-disable-next-line no-undef
```
juancarlosfarah, line 123: Same as above, but for `MessageChannel`.

juancarlosfarah (approved):
Looks good! Just a couple questions about ESLint.

Just to confirm that we're going directly to Yarn 3, right? Also, not sure why this file (`.yarn/releases/yarn-berry.cjs`) is different than one of the other repos.

juancarlosfarah (approved):
Looks good! Just a couple questions about ESLint.

Just to confirm that we're going directly to Yarn 3, right? Also, not sure why this file (`.yarn/releases/yarn-berry.cjs`) is different than one of the other repos.

pyphilia:
@juancarlosfarah I updated the PR so it works yarn3. I also updated the way tests run (using jest instead of `react-script`). Lots of files have changed because of linter.

package.json
```diff
@@ -6,6 +6,9 @@
   "contributors": [
     "Juan Carlos Farah"
   ],
+  "workspaces": [
+    "example"
```
juancarlosfarah, line 10: Why do we need this?

package.json
```diff
@@ -39,26 +43,31 @@
...
     "@typescript-eslint/eslint-plugin": "^2.26.0",
     "@typescript-eslint/parser": "^2.26.0",
     "babel-eslint": "^10.0.3",
+    "babel-jest": "^24.9.0",
     "cross-env": "^7.0.2",
     "enzyme": "3.11.0",
     "enzyme-adapter-react-16": "1.15.5",
     "enzyme-to-json": "3.6.1",
-    "eslint": "^6.8.0",
+    "eslint": "^6.6.0",
```
juancarlosfarah, line 70: Can we pin this and all the other "unpinned" libraries?

package.json
```diff
@@ -6,6 +6,9 @@
   "contributors": [
     "Juan Carlos Farah"
```
juancarlosfarah, line 7: Please add the contributors.

juancarlosfarah (changes requested):
Looks great! Just some minor comments. 🎉

package.json
pyphilia: It actually ease the process, as it `yarn` the workspaces at the same time as the core repository.

package.json
pyphilia: No, it depends on `react-scripts`. I'll see for the other packages. Perhaps it will rather be in a update dep branch.

juancarlosfarah: OK

juancarlosfarah:
Got it.

—
Juan Carlos Farah
Vice President of Research
[Graasp Association](https://graasp.org)

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #43, opened 2021-08-05
fix: create ref outside of constructor in AppItem

### LINKED ISSUE #42: Must call super constructor in derived class before accessing 'this' or returning from derived constructor
Related to AppItem.

pyphilia:
close #42

### DIFF
--- a/src/items/AppItem.js
+++ b/src/items/AppItem.js
@@ -42,10 +42,7 @@ class AppItem extends Component {
     url: null,
   };
 
-  constructor(props) {
-    super(props);
-    this.iframeRef = React.createRef();
-  }
+  iframeRef = React.createRef();
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/items/AppItem.js
```diff
@@ -42,10 +42,7 @@ class AppItem extends Component {
     url: null,
   };
 
-  constructor(props) {
-    super(props);
-    this.iframeRef = React.createRef();
-  }
+  iframeRef = React.createRef();
```
juancarlosfarah, line 45: Can you please add a comment regarding why we need this just for posterity? Thanks!

juancarlosfarah (changes requested):

pyphilia:
This actually is not working well with the ecosystem. The related bug seemed to have disappear. I close this PR for now.

PR #54, opened 2021-11-24
refactor: get dynamic height for app, pdf and links

### LINKED ISSUE #53: Adapt app and pdf iframe sizes
`document.window.innerHeight` after waiting for the loading might give the perfect height for pdf / apps.

Otherwise we should set it in the app template.

pyphilia:
close #53


pyphilia:
ref https://github.com/graasp/graasp-perform/issues/42

Julien-Torrent (approved):

PR #76, opened 2022-01-12
feat: add item flag menu to UI component

louisewang1:


src/ItemFlagMenu/ItemFlagMenu.tsx
```diff
@@ -0,0 +1,92 @@
...
+      overflow: 'auto',
+      maxHeight: FLAG_LIST_MAX_HEIGHT,
+    },
+    listTitle: {
+      fontSize: 'small',
+    },
+    flagItemButton: {
+      color: 'red',
+    },
+  }));
```
pyphilia, line 43: If there's no dependency with the component, you can define it outside.

src/ItemFlagMenu/ItemFlagMenu.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { FLAG_LIST_MAX_HEIGHT } from '../constants';
+import { Flag } from '../types';
+
+export interface ItemFlagMenuProps {
+  flags: Flag[];
+  onFlag: any;
+  open: boolean;
+  setOpen: any;
+  selectedFlag: Flag;
+  setSelectedFlag: any;
```
pyphilia, line 20: Could you try removing these `any` types? At worst go with `Function` and disable eslint.

pyphilia (changes requested):
NIce PR! I would maybe call the component ItemFlagDialog?

pyphilia (approved):

PR #83, opened 2022-01-17
feat: add FlagItemButton component

### LINKED ISSUE #82: Add flag item button to UI component

louisewang1:
close #82


src/ItemFlag/FlagItemButton.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FC } from 'react';
+import { useTranslation } from 'react-i18next';
+import { IconButton, Tooltip } from '@material-ui/core';
+import ReportIcon from '@material-ui/icons/Report';
+
+export interface FlagItemButtonProps {
+  setOpen: Function;
```
pyphilia, line 7: Would this work if you use `(arg: boolean) => void`?

src/ItemFlag/FlagItemButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const { t } = useTranslation();
+
+  const openItemFlagDialog = (): void => {
+    setOpen(true);
+  };
+
+  return (
+    <Tooltip title={t('Report')}>
+      <IconButton color='primary' onClick={openItemFlagDialog}>
+        <ReportIcon fontSize='large' />
```
pyphilia, line 20: Are other buttons using `large`? I would add it in the props, and set a default value. Same for `primary`.

pyphilia (changes requested):
Really nice!

pyphilia (approved):

PR #84, opened 2022-01-19
refactor: show button under iframe for manual opening

### LINKED ISSUE #81: Open link manually button disappeared

pyphilia:
This button was buggy, so I changed for a regular button under the iframe. This is also better for accessibility.

close #81


src/items/LinkItem.tsx
```diff
@@ -60,12 +46,12 @@ const LinkItem: FC<LinkItemProps> = ({
   saveButtonId,
   editCaption = false,
   showCaption = true,
-  loadingMessage = 'Link is Loading.',
-  openLinkMessage = 'Click here to open link manually.',
+  loadingMessage = 'Link is Loading...',
+  openLinkMessage = 'Click here to open the link manually',
   height: defaultHeight,
+  errorMessage = 'The link is malformed.',
```
juancarlosfarah, line 52: Are these messages hard-coded? Or can they be overridden when used? It's probably better if everything can be overridden.

src/items/LinkItem.tsx
```diff
@@ -105,6 +95,10 @@ const LinkItem: FC<LinkItemProps> = ({
     // }
   };
 
+  const onClick = (): void => {
+    redirect(url, { openInNewTab: true });
+  };
+
   const component = (
     <React.Fragment>
```
juancarlosfarah, line 103: I would import { Fragment } from React. Looks nicer and we're already importing things.

juancarlosfarah (approved):
Approving with some minor comments / questions! 🪨

src/items/LinkItem.tsx
pyphilia: These are default values, so yes they can be overridden!

juancarlosfarah (approved):
👍

PR #96, opened 2022-03-04
feat: add item id in post message keys

### LINKED ISSUE #95: Create dedicated channels for apps
This will avoid concurrency problems when multiple apps are opened at the same time in the same tab.

pyphilia:
This PR adds channels for postmessages. This fixes the problem of concurrency if two apps runs in the same tab.

close #95


spaenleh (approved):
Nothing to add, looks good 🛫

PR #98, opened 2022-03-08
fix: update TextEditor content when initialValue changes

alvrba:


pyphilia (approved):
Thanks for this PR! Don't forget to link the issue to this PR (and close the issue if it doesn't do by itself..)

PR #109, opened 2022-03-30
chore: add React memo to items components

alvrba:
Add React memo to:
- FileItem (that renders FileImage, FileAudio, FileVideo, and FilePDF)
- AppItem
- LinkItem


pyphilia (approved):
Cool! You might need to update the yarn.lock again.

PR #118, opened 2022-04-26
fix: remove error message when loading files

alvrba:


src/items/FileItem.tsx
```diff
@@ -52,7 +53,11 @@ const FileItem: FC<FileItemProps> = ({
       if (content) {
         // Build a URL from the file
         const fileURL = URL.createObjectURL(content);
-        setUrl(fileURL);
+        if (fileURL) {
+          setUrl(fileURL);
+        } else {
+          setUrl("error");
```
pyphilia, line 59: Factor out the string, since you reuse it below.

src/items/S3FileItem.tsx
```diff
@@ -51,7 +52,11 @@ const S3FileItem = ({
       if (content) {
         // Build a URL from the file
         const fileURL = URL.createObjectURL(content);
-        setUrl(fileURL);
+        if (fileURL) {
+          setUrl(fileURL);
+        } else {
+          setUrl("error");
```
pyphilia, line 58: Same as above.

pyphilia (approved):
Easy change 💾  I've added a small comment, you can merge after solving it!

PR #122, opened 2022-05-05
feat: add Collapse component

alvrba:


src/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,87 @@
...
+      },
+    },
+    expanded: {},
+    rootAccordionSummary: {
+      borderBottom: '1px solid rgba(0, 0, 0, .125)',
+      marginBottom: -1,
+      minHeight: 56,
+      flexDirection: 'row-reverse',
+      '&$expanded': {
+        minHeight: 56,
```
pyphilia, line 37: Factor out the height.

pyphilia (changes requested):
I would have love a screenshot, but it looks all good to me 🦓 I've left minor request changes.

PR #129, opened 2022-05-31
128/AddExtensionFeatureToIframesLinksAndApps

alvrba:


src/items/LinkItem.tsx
```diff
@@ -110,18 +134,20 @@ const LinkItem: FC<LinkItemProps> = ({
       </div>
       <div
         hidden={isLoading}
-        className={classes.iframeContainer}
-        style={{ height: height }}
+        className={clsx(
+          classes.iframeContainer,
+          !isExtendable && classes.iframeContainerWithoutExtension,
```
pyphilia, line 139: Here you can use another syntax which allows conditions:
```js
clsx(
    someclass,
    { [classes.xxx]: !isExtendable }
)
```

src/items/withExtension.tsx
```diff
@@ -0,0 +1,53 @@
...
+                  height: height,
+                  x: 0,
+                  y: 0,
+                }}
+                minHeight={IFRAME_MIN_HEIGHT}
+                resizeHandleComponent={{ bottom: <ExtensionIcon /> }}
+                resizeHandleStyles={{ bottom: resizeHandleStyles }}
+              >
+                {component}
+              </Rnd>
```
pyphilia, line 42: Could we refactor so we use makeStyles?

pyphilia (changes requested):
Nice PR, actually what about keeping the naming "collapse" instead of extendable? I think they are two opposites to describe the same feature. What do you think?

src/icons/ExtensionIcon.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import Box from '@material-ui/core/Box';
+import { makeStyles, createStyles, Theme } from '@material-ui/core/styles';
+import UnfoldMoreIcon from '@material-ui/icons/UnfoldMore';
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    icon: {
+      backgroundColor: theme.palette.grey[300],
+      fontSize: '24px',
```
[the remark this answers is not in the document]
alvrba: Apparently the prop for the size is fontSize (https://v4.mui.com/components/icons/#size)

src/items/withExtension.tsx
alvrba: I have tried to apply `makeStyles`, but it cannot override the default style. I will ask you tomorrow.

src/items/LinkItem.tsx
```diff
@@ -50,6 +58,7 @@ const LinkItem: FC<LinkItemProps> = ({
   openLinkMessage = 'Click here to open the link manually',
   height: defaultHeight,
   errorMessage = 'The link is malformed.',
+  isExtendable = false,
```
pyphilia, line 61: So `isExtendable` is `isResizable`? I think I'm still confused but could we also match the naming with `resizing`? This comment applies to all other files.

src/items/LinkItem.tsx
```diff
@@ -20,13 +22,17 @@ interface LinkItemProps {
...
   errorMessage?: string;
+  isExtendable?: boolean;
 }
 
 const useStyles = makeStyles((theme) => ({
   iframe: {
     width: '100%',
     border: 'none',
+  },
+  iframeWithoutExtension: {
```
pyphilia, line 33: `iframeWithoutResizer`?

pyphilia (changes requested):
Thanks for the changes, I hope we will agree on the naming^^

PR #131, opened 2022-06-02
130/redirect

### LINKED ISSUE #130: Add Authorization and RedirectionContent components

### LINKED ISSUE #132: Capitalize navigation component item menu

pyphilia:
This PR contains new components:
- withAuthorization: this checks whether the member is signed in
- RedirectionContent: content to show on redirection
- UserSwitch: button and menu to show the current user and the saved sessions 

close #130 
close #132


src/UserSwitch/UserSwitch.tsx
```diff
@@ -0,0 +1,278 @@
...
+  MouseEventHandler,
+  ReactElement,
+  useState,
+} from 'react';
+import {
+  isPseudonymizedMember,
+  isSessionExpired,
+  isError,
+} from '@graasp/utils';
+import { makeStyles } from '@material-ui/core/styles';
```
dialexo, line 13: Note for the future: we probably want to use the prettier-sort-imports plugin with a group dedicated to material / mui 😅

src/UserSwitch/UserSwitch.tsx
```diff
@@ -0,0 +1,278 @@
...
+  buildMemberMenuItemId,
+  member = Map(),
+  members = [],
+  isMemberLoading = false,
+  onMemberClick = () => () => undefined,
+  seeProfileText = 'See Profile',
+  signedOutTooltipText = 'You are not signed in.',
+}) => {
+  const classes = useStyles();
+  const [anchorEl, setAnchorEl] = useState<(EventTarget & Element) | null>(
```
dialexo, line 94: Consider preferring [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref) to `useState` if applicable? If I understood correctly, you only use it to locate the root `Box`. In which case you probably won't need `handleClick` anymore. I would find it more readable if you track the open state separately as a boolean in `const [isOpen, setOpen] = useState(false)` for instance. Let me know if this is not possible.

dialexo (changes requested):
Nice! 🎉 I've requested one major change regarding hook usage, and left some small comments. Let me know it they make sense 😛

example/package.json
```diff
@@ -25,9 +26,11 @@
     "@types/node": "link:../node_modules/@types/node",
     "@types/react": "link:../node_modules/@types/react",
     "@types/react-dom": "link:../node_modules/@types/react-dom",
+    "@types/react-router-dom": "link:../node_modules/@types/react-router-dom",
```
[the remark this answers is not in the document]
pyphilia: Yes it's the example repo, so it reuses the main module packages.

src/Authorization/RedirectionContent.tsx
```diff
@@ -0,0 +1,59 @@
...
+}) => {
+  const classes = useStyles();
+  return (
+    <Container id={id} className={classes.container}>
+      <GraaspLogo height={100} className={classes.logo} />
+      <div className={classes.text}>
+        <Typography variant='h4' align='center'>
+          {redirectionText ?? 'You are being redirected…'}
+        </Typography>
+        <Link to={link} className={classes.link}>
```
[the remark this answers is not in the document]
pyphilia: True, I've changed it to include external links.

src/UserSwitch/UserSwitch.tsx
pyphilia: Nice comment, but there's some issues here:
- This is actually a copy paste of mui's implementation of the menu. It doesn't change in the new version, so I guess it's safe to use it like this.
- An implementation with ref would work. But since the button can be custom, I cannot really set the ref on it. The wrapper component is a Box and I cannot set ref to it.
So I think I'll let it like that.

dialexo (approved):

PR #137, opened 2022-06-09
136/FixPdfVisualizationWithCollapse

alvrba:


src/items/FilePdf.tsx
```diff
@@ -24,25 +24,14 @@ const FilePdf: FC<FilePdfProps> = ({
...
   const classes = useStyles();
-  const [height, setHeight] = useState<number | string>(
-    defaultHeight ?? '100%',
-  );
-
-  const onLoad: ReactEventHandler<HTMLEmbedElement> = (e) => {
-    // set pdf height -> probably very high
-    const newHeight = (e.target as HTMLEmbedElement)?.offsetParent
-      ?.scrollHeight;
-    newHeight && setHeight(newHeight);
```
pyphilia, line 35: Does your implementation dynamically set the height? I think since you're wrapping the pdf in a new container, it doesn't get the right parent for the height. Would it be possible to get a height from the window maybe instead?

pyphilia (changes requested):
I've left a comment, I think we will miss the dynamic height part if you remove this code. See the comment below.

src/items/FilePdf.tsx
```diff
@@ -29,10 +31,14 @@ const FilePdf: FC<FilePdfProps> = ({
...
 
   const onLoad: ReactEventHandler<HTMLEmbedElement> = (e) => {
-    // set pdf height -> probably very high
-    const newHeight = (e.target as HTMLEmbedElement)?.offsetParent
-      ?.scrollHeight;
-    newHeight && setHeight(newHeight);
+    // only set pdf height if not using collapse
+    if (!showCollapse) {
+      // set pdf height -> probably very high
+      console.log((e.target as HTMLEmbedElement)?.offsetParent);
```
pyphilia, line 37: Remove console.

pyphilia (approved):

PR #139, opened 2022-06-13
Use PDFjs viewer

alvrba:


src/items/FilePdf.tsx
```diff
@@ -41,11 +41,13 @@ const FilePdf: FC<FilePdfProps> = ({
     }
   };
 
+  const src = `https://pdfviewer.dev.graasp.org/pdf/web/viewer.html?file=${url}`;
```
dialexo, line 44: Do you perform checks on the receiving end to prevent injection? Also, should this always be served from the dev domain?

src/items/FilePdf.tsx
```diff
@@ -41,11 +41,13 @@ const FilePdf: FC<FilePdfProps> = ({
...
+
   return (
     <embed
```
dialexo, line 47: I'm curious why this is not integrated as an iframe? We would also get security benefits if we tighten the cookie policy or simply serve from another domain.

dialexo (commented):
Not sure about the implementation. We should probably refine the integration policy (domain and HTML node type).

alvrba:
I close this PR because it is integrated in https://github.com/graasp/graasp-ui/pull/160

PR #144, opened 2022-06-24
refactor: add component to Button

### LINKED ISSUE #143: Add component prop to Button

pyphilia:
close #143

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -19,6 +19,7 @@ interface GraaspButtonProps {
   endIcon?: ReactNode;
   autoFocus?: boolean;
   fullWidth?: boolean;
+  component?: React.ElementType;
 }
 const useStyles = makeStyles((theme) => ({
   button: {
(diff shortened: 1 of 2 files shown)


dialexo (approved):
LGTM 🦄

src/Table/Table.tsx
```diff
@@ -104,10 +104,15 @@ const GraaspTable: React.FC<Props> = ({
...
 }) => {
+  const [rowData, setRowData] = useState(initialData);
   const [gridApi, setGridApi] = useState<GridApi>();
   const [selected, setSelected] = useState<string[]>([]);
   const classes = useStyles({ tableHeight });
 
+  useEffect(() => {
+    setRowData(initialData);
+  }, [initialData]);
+
```
dialexo, line 115: Is this effect required if you already set the initial `rowData`? Or does this force a re-render?

pyphilia: Yeah we need to force a re-render, because if you stay on the same component but the prop changed (in my memberships, I delete a membership from the parent, it should propagate but) then it doesn't update.

PR #148, opened 2022-07-04
145/navigation props

### LINKED ISSUE #145: Add color prop to Navigation

louisewang1:
close #145


src/Navigation/Navigation.tsx
```diff
@@ -88,15 +89,24 @@ interface NavigationProps {
...
   hostMap = {},
   id,
+  buttonColor = 'secondary',
+  buttonStyle,
+  triangleStyle,
 }) => {
   const classes = useStyles();
 
+  const button = buttonStyle || classes.button;
+  const triangle = triangleStyle || classes.triangle;
```
pyphilia, line 108: I don't think `xxxStyle` is a good name. Classname would be better.
Also, you can use `clsx` to merge classes. So you can still use the default class and apply yours on top of it.

```
clsx( classes.button, buttonClassName )
```

pyphilia (changes requested):
You probably need to rebase this branch before merging!
I've left some comments, let me know if you have questions 📽️

PR #160, opened 2022-08-30
157/DownloadAndUseFilesDirectlyFromAws

alvrba:


src/constants.ts
```diff
@@ -95,3 +95,11 @@ export const DEFAULT_LOADER_SIZE = 20;
 export const DRAG_ICON_SIZE = 18;
 
 export const HEADER_USERNAME_MAX_WIDTH = 120;
+
+export const PDF_VIEWER_LINK =
+  'https://assets.dev.graasp.org/pdf-viewer/web/viewer.html?file=';
```
dialexo, line 100: Should we use an environment variable passed as parameter instead?

src/constants.ts
```diff
@@ -95,3 +95,11 @@ export const DEFAULT_LOADER_SIZE = 20;
...
+
+export const SERVICE_TYPES = {
```
dialexo, line 102: Can this be grabbed from the SDK?

src/items/FilePdf.tsx
```diff
@@ -41,11 +43,13 @@ const FilePdf: FC<FilePdfProps> = ({
     }
   };
 
+  const urlWithPdfViewer = `${PDF_VIEWER_LINK}${encodeURIComponent(url)}`;
```
dialexo, line 46: We should probably perform some sanitization here. Maybe we could ensure that url starts with our domain?
Validation should be performed on the PDF viewer side, to ensure that malicious users cannot load arbitrary content or use our PDF viewer as a CDN.

src/items/FilePdf.tsx
```diff
@@ -41,11 +43,13 @@ const FilePdf: FC<FilePdfProps> = ({
...
+
   return (
     <embed
```
dialexo, line 49: Is there a reason why we use an embed and not an iframe? If we provide an external domain, we can also add additional security such as sandboxing.

dialexo (commented):
Thank you for this PR! I've left a few comments regarding architectural decisions, tagging @pyphilia

PR #208, opened 2023-01-19
fix: revert to default iframe

dialexo:
H5P cannot be displayed anymore since last update of H5P iframes (the origin of the iframe is now null because of the sandbox)


spaenleh (approved):
💾 LGTM !

PR #216, opened 2023-02-03
feat: add PlatformSwitch component

### LINKED ISSUE #215: Add PlatformSwitch component
This component will replace the current `Select` used to switch between the different platforms (Builder, Player, Library). Use the [`Pagination`](https://mui.com/material-ui/react-pagination/) MUI component and its `hideNextButton`, `hidePrevButton` and `renderItem` props to customize each button.

dialexo:
Closes #215 
Also updates Storybook and Typescript


src/PlatformSwitch/hooks.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export function usePlatformNavigation(hostsMapper: HostsMapper) {
+  return (platform: Platform) => {
+    // detect current item given url /[optionalPrefixA/optionalPrefixB/.../]<itemId>[...anything]
+    // todo: use proper top-level item context?
+    const itemRegex =
+      /^\/(?:\w+\/)*([0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}).*/;
+    const match = window.location.pathname.match(itemRegex);
+    const itemId = match?.[1] ?? '';
+    const url = hostsMapper[platform]?.(itemId) ?? '#';
```
spaenleh, line 54: Out of curiosity what does the `#` do ?

spaenleh (approved):
🐨 Looks good ! Thank you for the well-commented code, looking forward to using this !

src/PlatformSwitch/PlatformSwitch.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+    Library: {
+      onClick: () => alert('You clicked on Library'),
+    },
+    Analytics: {
+      disabled: true,
+      // this onClick should not trigger when disabled: true
+      onClick: () => alert('You clicked on Analytics'),
+    },
+  },
+};
```
pyphilia, line 56: You could add tests, if you really want (check your buttons are clickable) https://storybook.js.org/docs/react/writing-tests/interaction-testing

src/PlatformSwitch/hooks.ts
```diff
@@ -0,0 +1,57 @@
+import { redirect } from '@graasp/sdk';
+
+/** Enumeration of available platforms */
+export enum Platform {
+  Builder = 'Builder',
+  Player = 'Player',
+  Library = 'Library',
+  Analytics = 'Analytics',
+}
```
pyphilia, line 9: We have `Context` but I guess context is not a clear wording. Should we match one with the other?

src/PlatformSwitch/hooks.ts
pyphilia: Would it make sense to get the id as a prop instead? Each platform can easily get the id with `useParams()`

src/PlatformSwitch/hooks.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Generates a default hosts mapper given a record of platform to hostname
+ * Only the origin part of the given hostname will be used!
+ *
+ * The default mapping is:
+ *  BUILDER_HOST/items/<itemId>
+ *  PLAYER_HOST/<itemId>
+ *  LIBRARY_HOST/<itemId>
+ *
```
pyphilia, line 22: For analytics it is also `<host>/items/<id>`

src/PlatformSwitch/hooks.ts
```diff
@@ -0,0 +1,57 @@
...
+  return Object.fromEntries(
+    Object.entries(hostsUrls).map(([platform, url]) => {
+      const origin = new URL(url).origin;
+      // in Builder platform, should appear under /items
+      const prefixIfBuilder = platform === Platform.Builder ? 'items/' : '';
+      return [
+        platform,
+        (itemId: string) => `${origin}/${prefixIfBuilder}${itemId}`,
+      ];
+    }),
```
pyphilia, line 37: Actually there's util function to build a builder link https://github.com/graasp/graasp-sdk/blob/80bd60fb83e2b5fe3721a975184f99265b1dae66/src/utils/navigation.ts#L81
I didn't know where to set the responsibility of defining the url. But since we used it at many places so it was simpler to reuse this bit of code.

pyphilia (approved):
I've added suggestion and minor comments, thanks for the PR!

src/PlatformSwitch/hooks.ts
dialexo: > Out of curiosity what does the `#` do ?

It redirects to "nowhere", it's basically a no-op when redirecting in a browser while keeping the current URL.
If you prepend some identifier afterwards, it will be bound to the first element which has the same id, effectively scrolling to the corresponding part of the page. E.g. `<a href="#news">` will let the browser stay on the same page, and jump to the news section, given that some `<div id="news">` exists.

src/PlatformSwitch/PlatformSwitch.stories.tsx
dialexo: I think the interactions are simple enough, given the UI component only exposes `onClick`s. The integrations tests will make more sense IMO.

src/PlatformSwitch/hooks.ts
dialexo: True, I though they were defined in each platform codebase initially. I think renaming `Context` to `Platform` is more precise. I think it would break a bunch of stuff though.

src/PlatformSwitch/hooks.ts
dialexo: Aren't the analytics currently hosted under Builder?

src/PlatformSwitch/hooks.ts
dialexo: I think we have a weird situation where each platform knows best what link is relevant for a given item ID, but another platform may want to use a specific link for a given functionality of another platform. Also in this case, we should be using `buildItemLinkForBuilder` only if the platform is Builder, otherwise the route is different. 

What about having platforms expose some kind of "default" route for any itemId, that would show the relevant "home" page for that id? Not sure where we should put this either.

src/PlatformSwitch/hooks.ts
dialexo: Update: i've tried using `Context` but all things break since I'm depending on the value of the enum. I could be rewriting the list without the deprecations, but it would be the same as using `Platform`. I'll stick with this and open an issue in  `graasp-sdk` to unify in the future.

PR #224, opened 2023-02-08
fix: clear hover state when platform opens in new tab

dialexo:
Fixes a bug where the hover state could not be cleared when a platform was opened in a new tab


pyphilia (approved):
LGTM 🎣

Issue #250, opened 2023-02-21
Add SVG to supported image items


swouf:
The mimetype is `image/svg+xml`.

spaenleh:
Fixed in #256

PR #256, opened 2023-02-22
feat: support wav as audio format

### LINKED ISSUE #255: Add `audio/wav` and `audio/x-wav` to supported audio MIMETYPEs
Ref https://github.com/graasp/graasp-player/issues/193

spaenleh:
Currently audio files in WAV format are shown as the default file (download button)  while they could be played in the browser.

Adding the MIME type will allow the files to be displayed as an audio file and played directly.

closes #255

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -4,7 +4,7 @@ export const DRAWER_HEADER_HEIGHT = 55;
 export const LOADING_TEXT = '…';
 
 export const MIME_TYPES = {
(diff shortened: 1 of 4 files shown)


spaenleh:
@swouf should we also include `image/svg+xml` to fix #250 ?

swouf:
I added `image/svg+xml` in the mimetype list. I also added a story.

closes #250

dialexo (approved):

PR #259, opened 2023-02-23
fix: improve card component and add stories

### LINKED ISSUE #258: Improve `Card`
Improve Layout of card and make it more flexible.
- Allow text to get closer to action button.
- "More" actions button should not move depending on whether there is a creator or not
- Description should be able to occupy as much space as possible and signal overflow
- image should keep a square format whatever the ratio of the card
  - In case the card is too tall, the image should not occupy more than half the card.
- Simplify the layout using `Stack` instead of `Grid` elements

spaenleh:
This PR improves the card component.

card-improvements-description-all

closes #258


pyphilia (approved):
LGTM 🐘

PR #264, opened 2023-03-01
fix: set default skeleton variant to circular for Avatar

### LINKED ISSUE #263: Set default for skeleton variant in Avatar

spaenleh:
This PR updates the default skeleton shape value for the `Avatar` component.

closes #263

### DIFF
--- a/src/items/ItemSkeleton/ItemSkeleton.tsx
+++ b/src/items/ItemSkeleton/ItemSkeleton.tsx
@@ -54,7 +54,7 @@ const ItemSkeleton: React.FC<ItemSkeletonProps> = ({
         ItemType.S3_FILE,
         ItemType.LINK,
         ItemType.APP,
-      ] as string[]
+      ] as `${ItemType}`[]
     ).includes(itemType): {
       return (
         <Skeleton
--- a/src/items/ItemSkeleton/ItemSkeleton.tsx
+++ b/src/items/ItemSkeleton/ItemSkeleton.tsx
@@ -16,7 +16,7 @@ export interface ItemSkeletonProps {
    */
   isChildren: boolean;
   isCollapsible?: boolean;
-  itemType: ItemType;
+  itemType: `${ItemType}` | ItemType;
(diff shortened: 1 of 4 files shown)


src/items/ItemSkeleton/ItemSkeleton.tsx
```diff
@@ -16,7 +16,7 @@ export interface ItemSkeletonProps {
    */
   isChildren: boolean;
   isCollapsible?: boolean;
-  itemType: ItemType;
+  itemType: `${ItemType}` | ItemType;
```
dialexo, line 19: Cool 😎

src/items/ItemSkeleton/ItemSkeleton.tsx
```diff
@@ -48,12 +48,14 @@ const ItemSkeleton: React.FC<ItemSkeletonProps> = ({
...
-      ItemType.APP,
-    ].includes(itemType): {
+    case (
+      [
+        ItemType.LOCAL_FILE,
+        ItemType.S3_FILE,
+        ItemType.LINK,
+        ItemType.APP,
+      ] as string[]
+    ).includes(itemType): {
```
dialexo, line 58: Why is this typing necessary?

dialexo (approved):

src/items/ItemSkeleton/ItemSkeleton.tsx
spaenleh: Without it, the type of the array is `ItemType[]` and the values of the union type can't be compared because the types do not match.
Capture d’écran 2023-03-02 à 15 05 42

spaenleh: Maybe we can do:
```ts
[
        ItemType.LOCAL_FILE,
        ItemType.S3_FILE,
        ItemType.LINK,
        ItemType.APP,
      ] as `${ItemType}`[]
    ).includes(itemType):
```

PR #268, opened 2023-03-01
267 improve optional types on itemloginauthorization

spaenleh:
- fix: optional types on ItemLoginSchema
- fix: type of withCollapse comp uses ItemRecord

### DIFF
--- a/src/itemLogin/ItemLoginAuthorization.tsx
+++ b/src/itemLogin/ItemLoginAuthorization.tsx
@@ -19,7 +19,7 @@ import ItemLoginScreen, { SignInPropertiesType } from './ItemLoginScreen';
 export type ItemLoginAuthorizationProps = {
   signOut: () => void;
   signIn: (args: { itemId: string } & SignInPropertiesType) => void;
-  itemId?: UUID;
+  itemId: UUID;
   useCurrentMember: () => UseQueryResult<MemberRecord>;
   useItem: (itemId?: string) => UseQueryResult<ItemRecord>;
   useItemLogin: (itemId?: string) => UseQueryResult<ItemLoginRecord>;
--- a/src/itemLogin/ItemLoginAuthorization.tsx
+++ b/src/itemLogin/ItemLoginAuthorization.tsx
@@ -19,10 +19,10 @@ import ItemLoginScreen, { SignInPropertiesType } from './ItemLoginScreen';
 export type ItemLoginAuthorizationProps = {
   signOut: () => void;
   signIn: (args: { itemId: string } & SignInPropertiesType) => void;
-  itemId: UUID;
+  itemId?: UUID;
   useCurrentMember: () => UseQueryResult<MemberRecord>;
-  useItem: (itemId: string) => UseQueryResult<ItemRecord>;
-  useItemLogin: (itemId: string) => UseQueryResult<ItemLoginRecord>;
+  useItem: (itemId?: string) => UseQueryResult<ItemRecord>;
+  useItemLogin: (itemId?: string) => UseQueryResult<ItemLoginRecord>;
   Error?: ReactElement;
   memberIdInputId?: string;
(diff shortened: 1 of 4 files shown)


src/itemLogin/ItemLoginAuthorization.tsx
```diff
@@ -21,8 +21,8 @@ export type ItemLoginAuthorizationProps = {
   signIn: (args: { itemId: string } & SignInPropertiesType) => void;
   itemId: UUID;
   useCurrentMember: () => UseQueryResult<MemberRecord>;
-  useItem: (itemId: string) => UseQueryResult<ItemRecord>;
-  useItemLogin: (itemId: string) => UseQueryResult<ItemLoginRecord>;
+  useItem: (itemId?: string) => UseQueryResult<ItemRecord>;
+  useItemLogin: (itemId?: string) => UseQueryResult<ItemLoginRecord>;
```
dialexo, line 25: I guess this will simply return or error out if they are undefined. But should we enforce this?

src/itemLogin/ItemLoginScreen.tsx
```diff
@@ -107,9 +107,9 @@ const ItemLoginScreen: FC<ItemLoginScreenProps> = ({
 }) => {
   const { t } = useTranslation();
   const loginModeRef = useRef(null);
-  const [password, setPassword] = useState<string | undefined>(undefined);
-  const [username, setUsername] = useState<string | undefined>(undefined);
-  const [memberId, setMemberId] = useState<string | undefined>(undefined);
+  const [password, setPassword] = useState<string>();
+  const [username, setUsername] = useState<string>();
+  const [memberId, setMemberId] = useState<string>();
```
dialexo, line 112: Doesn't `useState` hold `undefined` by default if no argument is passed?
In which case the type is thus technically wrong: it should either be `useState<string | undefined>();` or `useState<string>("");` I think.

dialexo (changes requested):
I am not sure about the types here. Is the new behaviour desirable?

src/itemLogin/ItemLoginScreen.tsx
spaenleh: You are right that `useState` holds undefined as the default value. But they typed it so that you only pass the other type as the generic argument as can be seen below:
```ts
    /**
     * Returns a stateful value, and a function to update it.
     *
     * @version 16.8.0
     * @see https://reactjs.org/docs/hooks-reference.html#usestate
     */
    function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];
    // convenience overload when first argument is omitted
    /**
     * Returns a stateful value, and a function to update it.
     *
     * @version 16.8.0
     * @see https://reactjs.org/docs/hooks-reference.html#usestate
     */
    function useState<S = undefined>(): [S | undefined, Dispatch<SetStateAction<S | undefined>>];
```
So with `useState<string>()` you get a variable that is `string | undefined` and if you specify the value of the argument then it will be just the type of the argument.

Very good comment though, TIL how `useState` is typed !

src/itemLogin/ItemLoginAuthorization.tsx
spaenleh: This typing is more consistent with the typing of the query. But you are right the behavior of the query is to error with "Undefined Argument" when the value is `undefined`.

src/itemLogin/ItemLoginScreen.tsx
dialexo: You are right, I somehow got confused and thought this was equivalent to `const [state, setState]: string = useState<string>()` but this would obviously be `const [state, setState]: string | undefined = useState<string>()`, and anyway we let TS infer. My bad! 😕

spaenleh:
Is it good to merge then ?

dialexo (approved):

PR #274, opened 2023-03-09
fix: use MimeTypes and ThumbnailSize from sdk

### LINKED ISSUE #273: Use ThumbnailSize from sdk

spaenleh:
This PR uses sdk verion `0.8.0` and updates the usage of `MIME_TYPES` and `ThumbnailSize` to the ones provided by sdk.

closes #273

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -5,23 +5,6 @@ export const DEFAULT_DIRECTION = 'ltr';
 export const DRAWER_HEADER_HEIGHT = 55;
 export const LOADING_TEXT = '…';
 
-/**
- * @deprecated use from graasp/sdk
- */
-export const MIME_TYPES = {
-  IMAGE: ['image/png', 'image/jpg', 'image/gif', 'image/jpeg', 'image/svg+xml'],
-  VIDEO: [
-    'video/mp4',
-    'video/x-m4v',
-    'video/ogg',
-    'video/quicktime',
-    'video/webm',
-  ],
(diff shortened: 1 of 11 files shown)


src/constants.ts
```diff
@@ -1,8 +1,13 @@
+import { ThumbnailSize } from '@graasp/sdk/frontend';
+
 export const DRAWER_WIDTH = 240;
 export const DEFAULT_DIRECTION = 'ltr';
 export const DRAWER_HEADER_HEIGHT = 55;
 export const LOADING_TEXT = '…';
 
+/**
```
dialexo, line 8: Q: why can't we just remove this?

src/icons/ItemIcon.tsx
```diff
@@ -93,23 +94,23 @@ const ItemIcon: FC<ItemIconProps> = ({
     case ItemType.LOCAL_FILE:
     case ItemType.S3_FILE: {
       if (mimetype) {
-        if (MIME_TYPES.IMAGE.includes(mimetype)) {
+        if (MIME_TYPES.isImage(mimetype)) {
```
dialexo, line 97: Nitpick: not sure if it makes sense that `MIME_TYPES` is uppercase, it seems particularly striking here on a method call. I see it more as a static object, so I would probably write it as `MimeTypes`. Should this be changed in SDK?

dialexo (approved):
LGTM 🏖️ I've left 2 small nitpicks

src/icons/ItemIcon.tsx
spaenleh: Yes we can change the naming in sdk.

PR #283, opened 2023-03-14
282 display item badges on card

spaenleh:
This PR adds a new component `ItemBadges` that displays hidden, published, public and pinned status of items.

This Component can be used on the Card component as seen below:
Capture d’écran 2023-03-13 à 13 02 20


dialexo (approved):
Looks very nice thank you! :sparkle:

PR #289, opened 2023-03-15
195 adds collapse to h5p

### LINKED ISSUE #195: Add collapse to h5p
Add a prop to `H5PItem.tsx` called `showCollapse` to enable or disable the Collapse wrapper component.

Similar to this 
https://github.com/graasp/graasp-player/blob/780ff89e07fd23d7783c94e9751aead6091a2098/src/components/common/Item.js#L322 
but this behavior should happen in graasp-ui.

MalinSvenberg:
Closes #195 

Adds `showCollapse` as a prop to `H5PItem`, `AppItem`, `FileItem`, `DocumentItem` and `LinkItem`. Makes these items collapsable. 

I'm unsure if I maybe shouldn't force H5PItem to both have item and itemId as props?


spaenleh (changes requested):
Looks good, I made a comment to try and harmonize the way the components are returned.

Regarding the issue of passing the full item to H5P, I think we have 2 option:
- stay with what you did, passing the full item, as this is what the `withCollapse` function is expecting
- pass only needed props. In this I think the `withCollapse` doesn't need the full Item to only extract the name property. 

I am leaning towards the second option, as it would make the withCollapse component less dependent on the item.

what do you think @MalinSvenberg ? We can also ask @dialexo since he made the H5PIntegration and can give us advice on what would be best regarding props.

pyphilia:
You need to remove an unused import for build to work. Also your branch is a bit behind main so you will need to rebase, let me know if you don't know how to do it we can do it together tomorrow.

src/items/FileItem.tsx
```diff
@@ -138,33 +140,37 @@ const FileItem: FC<FileItemProps> = ({
...
+  } else {
+    fileItem = (
+      <DownloadButtonFileItem
+        id={id}
+        name={originalFileName}
+        url={url}
+        text={downloadText}
+      />
+    );
   }
```
pyphilia, line 167: This file was originally quite complex. 
Actually I would display the caption even tho there's a default or a download button.
Maybe since you have a complex if-else logic as well, I would refactor with a function like this (not sure how to call the function tho):
```javascript

const buildComponent = () => {

  if(mimetype) {
  // here mimetype switch
  return component;
  }
    if (defaultItem) {
    return defaultItem;
  }
  return <Download...>
}


fileItem = buildComponent( )

if(showCaption)...
if(showCollapse)...

```

src/items/H5PItem.tsx
```diff
@@ -1,15 +1,23 @@
 import * as uuid from 'uuid';
 
+import { WorkHistory } from '@mui/icons-material';
```
pyphilia, line 3: Remove this.

src/items/LinkItem.tsx
```diff
@@ -92,98 +98,108 @@ const LinkItem: FC<LinkItemProps> = ({
...
+  const getComponent = (): JSX.Element | null => {
+    // if available, display specific player
+    const html = extra?.html;
+    if (html) {
+      const component = (
+        <div id={id} dangerouslySetInnerHTML={{ __html: html }} />
+      );
+      if (showCaption) {
+        return CaptionWrapper(component);
+      }
```
pyphilia, line 110: I think we want to show the caption each time so it makes sense to apply it outside of the function, before show collapse.

src/items/LinkItem.tsx
```diff
@@ -92,98 +98,108 @@ const LinkItem: FC<LinkItemProps> = ({
...
+      redirect(url, { openInNewTab: true });
+    };
+
+    const renderIframe = (): JSX.Element | null => {
+      if (!showIframe) {
+        return null;
+      }
+
+      const iframe = (
+        <StyledIFrame
```
pyphilia, line 140: Don't nest functions, these can defined outside.

pyphilia (changes requested):
I've added my comments, I think there's still refactor to improve the readability (it wasn't that readable before either).
Let me know if you have questions.

MalinSvenberg:
I changed withCollapse so that it only takes in the item's name instead of the entire item, since that was the only thing it used from the item.

src/Collapse/withCollapse.tsx
```diff
@@ -1,18 +1,16 @@
...
-
 import Collapse from './Collapse';
 
 interface WithCollapseProps {
-  item: ItemRecord;
+  itemName: string;
 }
 
-function withCollapse({ item }: WithCollapseProps) {
+function withCollapse({ itemName }: WithCollapseProps) {
```
pyphilia, line 9: It might have been easier to use `name` so that the other components could use `withCollapse(item)`, just saying.

pyphilia (approved):
It's all good to me 🦒

spaenleh (approved):
Looks good ! Thank you for the changes, and sorry for the delay ⌛

PR #291, opened 2023-03-15
fix: search bar can be compressed and is responsive

spaenleh:
This PR updates the search bar to be able to compress (minWidth:0) and the 
`width` parameter can now be used to set a `maxWidth`. Otherwise he bar will always 
take the space available in its parent.


pyphilia (approved):
LGTM 🐉

PR #313, opened 2023-03-30
fix: add collapsible and chat to itemstatus

### LINKED ISSUE #309: Add collapsible status badge

spaenleh:
This PR adds the `isCollapsible` and `showCahtbox` statuses to the item badges.
Capture d’écran 2023-03-30 à 12 01 51

closes #309

### DIFF
--- a/src/ItemBadges/ItemBadges.tsx
+++ b/src/ItemBadges/ItemBadges.tsx
@@ -1,4 +1,10 @@
-import { Public, PushPin, VisibilityOff } from '@mui/icons-material';
+import {
+  Chat,
+  Public,
+  PushPin,
+  UnfoldLess,
+  VisibilityOff,
+} from '@mui/icons-material';
 import { Avatar, AvatarGroup, Tooltip } from '@mui/material';
 
 import React from 'react';
@@ -32,6 +38,10 @@ type Props = {
   isPublishedTooltip?: string;
   isPinned?: boolean;
   isPinnedTooltip?: string;
+  isCollapsible?: boolean;
+  isCollapsibleTooltip?: string;
+  showChatbox?: boolean;
+  showChatboxTooltip?: string;
   backgroundColor?: string;
 };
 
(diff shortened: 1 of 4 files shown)


src/ItemBadges/ItemBadges.tsx
```diff
@@ -44,10 +52,14 @@ const ItemBadges = ({
...
   isPublicTooltip = 'Public',
+  isCollapsible = false,
+  isCollapsibleTooltip = 'Collapsible',
+  showChatbox = false,
+  showChatboxTooltip = 'Chat',
   backgroundColor = 'grey.600',
 }: Props): JSX.Element => {
   return (
-    <AvatarGroup>
+    <AvatarGroup max={10}>
```
dialexo, line 62: Info: do we use this `max` prop as a "safety" to display properly?

src/ItemBadges/ItemBadges.tsx
```diff
@@ -44,10 +52,14 @@ const ItemBadges = ({
   isPublishedTooltip = 'Published',
   isPublic = false,
   isPublicTooltip = 'Public',
+  isCollapsible = false,
+  isCollapsibleTooltip = 'Collapsible',
+  showChatbox = false,
+  showChatboxTooltip = 'Chat',
```
dialexo, line 58: Should the tooltips get their values from translations?

dialexo (approved):
Thank you 🐱 I have some questions but LGTM

src/ItemBadges/ItemBadges.tsx
spaenleh: Yes I set it to an arbitrary number higher than the number of badges we use, the default in MUI is `5` after which the avatars are grouped and displayed as (+X)

src/ItemBadges/ItemBadges.tsx
spaenleh: Yes the consumers of this component should set the tooltips with translated content. The default values are here in case this is used without translations. Maybe I should make the tooltip texts mandatory and not provide default values ?

dialexo: I think it's fine for now, but yes IMO the boundaries are not well-defined. The consumers still have to each redo the work for the translations in this instance, even though the semantics that are tied to the item models are enclosed from within the component.

PR #330, opened 2023-04-16
fix: export public constants and colocate local constants

spaenleh:
This PR exposes public constants and colocates local constants closer to where they are used.


pyphilia (approved):
LGTM :electron:

PR #331, opened 2023-04-17
fix: app iframe loading forever

spaenleh:


pyphilia (approved):
I believe in your refactor 🙏

src/items/AppItem.tsx
```diff
@@ -79,260 +68,109 @@ export interface AppItemProps {
...
+  permission = PermissionLevel.Read,
+  isResizable = false,
+  showCollapse,
+  onSaveCaption,
+  onCancelCaption,
+  saveButtonId,
+  cancelButtonId,
+}: AppItemProps): JSX.Element => {
+  // state
+  const [isiFrameLoading, setIsiFrameLoading] = useState(true);
```
dialexo, line 93: style nitpick (since the DOM element type is `HTMLIFrameElement`)
```suggestion
  const [isIFrameLoading, setIsIFrameLoading] = useState(true);
```

src/items/appItemHooks.ts
```diff
@@ -0,0 +1,143 @@
...
+    // receive message from app through MessageChannel
+    const onMessage = async (e: MessageEvent): Promise<void> => {
+      const { data, origin: requestOrigin } = e;
+
+      const POST_MESSAGE_KEYS = buildPostMessageKeys(item.id);
+
+      // responds only to corresponding app
+      if (!appUrl?.includes(requestOrigin)) {
+        return;
+      }
```
dialexo, line 59: We could check for `source` as well, to ensure that it comes from the iframe (https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#the_dispatched_event)

src/items/appItemHooks.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      const { type, payload } = JSON.parse(data);
```
dialexo, line 61: Note for the future: we may still want to validate this data to be sure

src/items/appItemHooks.ts
```diff
@@ -0,0 +1,143 @@
...
+              apiHost,
+              itemId: item.id,
+              settings: item.settings,
+              memberId: member?.id,
+              permission,
+              lang: item.settings?.lang || member?.extra?.lang || DEFAULT_LANG,
+              context,
+            },
+          }),
+          '*',
```
dialexo, line 132: This should definitely be set to the target app origin. See https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#syntax at `targetOrigin`. Can this be set to `new URL(appUrl).origin`?

spaenleh: I will try.

src/items/appItemHooks.ts
spaenleh: In my tests the `source` is `null`

src/items/appItemHooks.ts
spaenleh: It works, the messages target our domain now, i.e. `apps.dev.graasp.org`

src/items/appItemHooks.ts
dialexo: Ah right because of the sandboxing

dialexo (approved):
Nice work!

PR #340, opened 2023-05-02
feat: add tooltips to platform switch buttons

### LINKED ISSUE #339: Add tooltips to the Platform switch buttons

MalinSvenberg:
Closes #339

Adds tooltips to the platform switch. Right now the default value is taken from `Platform`, so it will say `Builder`, `Player`, `Library` and `Analytics` automatically. Maybe no tooltip is better as default?  

 This is how it would look in Builder.

https://user-images.githubusercontent.com/56155987/235690167-b9337921-6813-4598-9169-74c550743633.mov


src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -126,20 +138,25 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
...
-          display: 'flex',
-          cursor: platformProps?.disabled ? 'default' : 'pointer',
-        }}
-        {...mouseHoverEvents}
-        onClick={platformProps?.disabled ? undefined : platformProps?.onClick}
-        onMouseDown={
-          platformProps?.disabled ? undefined : platformProps?.onMouseDown
-        }
+      <Tooltip
+        title={platformProps?.disabled ? undefined : title}
```
pyphilia, line 142: Isn't it possible to wrap with the tooltip component only if the platform is not disabled?

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -149,6 +166,8 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
       // the last icon does not need margin at the end
       sx={{ mr: index === platforms.length - 1 ? 0 : spacing }}
       platform={platform}
+      title={tooltips ? tooltips[index] : undefined}
+      placement={placements ? placements[index] : undefined}
```
pyphilia, line 170: I think I would go with objects instead of indexes. Eg. `tooltips[platform]` and `placements[platform]`

pyphilia (changes requested):
Thanks for your changes, could you also include some tests/examples in storybook? It would check the implementation works with some variety of props.

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -19,6 +20,10 @@ export type PlatformSwitchProps = {
   accentColor?: string;
   /** Color of the icons when the corresponding platform is disabled */
   disabledColor?: string;
+  /** Tooltips to add to the buttons, in order left to right */
+  tooltips?: string[];
+  /** Placements of tooltips, in order left to right */
+  placements?: TooltipProps['placement'][];
```
dialexo, line 26: To avoid passing indexed arrays, have a look at `platformProps` at line 32 below: it already provides you with a mapping from platform to props. Consider converting from arrays to a platform prop key here (simply add `tooltip?: string` and `placement?: TooltipProps['placement']`)

https://github.com/graasp/graasp-ui/blob/70cdc2916da4d38ff423a685e3aef0923e2d5b03/src/PlatformSwitch/PlatformSwitch.tsx#L32-L51

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -74,6 +79,9 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
   color = SECONDARY_COLOR,
   accentColor = PRIMARY_COLOR,
   disabledColor = '#CCC',
+  // if no custom tooltip, name from the platform is used
+  tooltips = Object.values(Platform),
+  placements,
```
dialexo, line 84: If you apply the previous suggestion, these will not be needed anymore

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -82,9 +90,13 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
...
     /** Platform which button should be rendered */
     platform: Platform;
+    /** Tooltip to be used for button */
+    title?: string;
+    /** Tooltip's placement*/
+    placement?: TooltipProps['placement'];
     /** Styles applied to the underlying icon */
     sx?: SxProps;
-  }> = ({ platform, sx }) => {
+  }> = ({ platform, title, placement, sx }) => {
```
dialexo, line 99: Similar as above

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -126,20 +138,25 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
```
dialexo, line 138: You could hence derive from platformProps / compute the defaults here

src/PlatformSwitch/PlatformSwitch.tsx
dialexo: I agree, see my above comments which would solve this.

src/PlatformSwitch/PlatformSwitch.tsx
dialexo: True though not sure which is cleaner

dialexo (changes requested):
Thank you for your work 👍 Joining what @pyphilia said, I think using the existing `platformProps` will solve your problems and make the code more elegant

dialexo (approved):
LGTM 🐭 Thank you!

PR #342, opened 2023-05-04
fix: use precise item type in DocumentItem given showCollapse

### LINKED ISSUE #341: Update type of `DocumentItem`
See https://github.com/graasp/graasp-builder/issues/614

The props must be changed to accept a discriminated union of `& ({showCollapse: true, item: { name: string, extra: DocumentItemExtra }} | {showCollapse: false, item: { extra: DocumentItemExtra }})`

Currently, the collapsible property is implemented by each item display (because some items may display differently if they are in a collapsible state, such as iframes). In the future, consider moving to a more decoupled API:

```tsx
<Collapsible>
   <FooItem />
</Collapsible>
```

if `FooItem` requires some changes because of `Collapsible`, use a context and provide a hook instead. This makes it possible to add Collapsible to anything, and only specify additional implementation when needed (instead of forcing the API to change at each item level).

```tsx
// Collapsible.tsx
const useIsCollapsible = () => useContext(collapsibleContext)
const Collapsible = () => {
    return <Context.Provider value={collapsibleContext} />
      //... implementation of collapsible
      {children}
    </Context.Provider>
}

// FooItem.tsx
const FooItem = () => {
  const collapsibleState = useIsCollapsible()
  return <>{ /* implementation that depends on collapsibleState */}</>
}
```

dialexo:
Closes #341

### DIFF
--- a/src/items/DocumentItem.tsx
+++ b/src/items/DocumentItem.tsx
@@ -2,19 +2,18 @@ import { Alert, Typography } from '@mui/material';
 
 import React, { FC } from 'react';
 
-import { getDocumentExtra } from '@graasp/sdk';
+import { DocumentItemType, getDocumentExtra } from '@graasp/sdk';
 import { DocumentItemTypeRecord } from '@graasp/sdk/frontend';
 
 import withCollapse from '../Collapse/withCollapse';
 import TextEditor from '../TextEditor';
 
-export interface DocumentItemProps {
+export type DocumentItemProps = {
   cancelButtonId?: string;
   cancelButtonText?: string;
   edit?: boolean;
   emptyMessage?: string;
   id?: string;
-  item: DocumentItemTypeRecord;
   maxHeight?: string | number;
   onCancel?: (text: string) => void;
   onChange?: (text: string) => void;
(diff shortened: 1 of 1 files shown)


src/items/DocumentItem.tsx
```diff
@@ -24,9 +23,11 @@ export interface DocumentItemProps {
   saveButtonText?: string;
   showActions?: boolean;
   showEmpty?: boolean;
-  showCollapse?: boolean;
   styles?: React.CSSProperties;
-}
+} & (
+  | { showCollapse: true; item: DocumentItemTypeRecord }
+  | { showCollapse?: false; item: Pick<DocumentItemType, 'extra'> }
```
spaenleh, line 29: I am curious whether it works if you don't have exactly the same type, since you are marking `showCollapse` as optional.

spaenleh (approved):
✅ This looks good to me if it works in the builder 🔨  !

src/items/DocumentItem.tsx
dialexo: Yes, for the compiler this is the same as `showCollapse: true | false | undefined`. So `undefined` is handled by the second type.

src/items/DocumentItem.tsx
```diff
@@ -44,7 +45,7 @@ const DocumentItem: FC<DocumentItemProps> = ({
   saveButtonText,
   showActions,
   showEmpty,
-  showCollapse = false,
+  showCollapse,
```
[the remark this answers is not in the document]
dialexo: Not sure, in any case since we handle it with `if(showCollapse)` both `false` and `undefined` are captured

PR #349, opened 2023-05-08
fix: display buttons in platform switch as links

### LINKED ISSUE #299: Render platform switch button as href'd link
This will support "open in new tab" and other interactions natively

dialexo:
Closes #299 and always redirect to library home as well


spaenleh (approved):
🦋 Looks good to me !

PR #363, opened 2023-06-09
362 add button role profile wrapper header

### LINKED ISSUE #362: Add role button to the profile wrapper in the header
This will make it possible for assistive technologies to know that it is clickable.

MalinSvenberg:
Closes #362 

Adds role button to the profile wrapper working as a button in the `UserSwitch`. As well as aria-attributes used for pop-up menus. 

Also adds this to `HeaderUserInformation` so this could be used in `UserSwitch` later (I didn't do it now since I can't test so everything works as it should in Builder etc). 

I've added a mock user of memberRecord type, however I don't know if this should be in fixtures. I didn't add any argTypes to the story, since I don't know what they should be. 

Could also be an idea to set an aria-hidden on the tooltip when the text for it is the username, since otherwise the username is read out twice by screen readers. However this would also hide the profile image from the accessibility tree, so it's a bit of a trade off.

### DIFF
--- a/src/HeaderUserInformation/HeaderUserInformation.tsx
+++ b/src/HeaderUserInformation/HeaderUserInformation.tsx
@@ -75,7 +75,7 @@ const HeaderUserInformation: FC<HeaderUserInformationProps> = ({
       id={id}
       role={role}
       aria-haspopup={popUpType}
-      aria-controls={popUpId}
+      {...(popUpId ? { 'aria-controls': popUpId } : {})}
       aria-expanded={isPopUpOpen}
     >
       <Tooltip
--- a/src/HeaderUserInformation/HeaderUserInformation.tsx
+++ b/src/HeaderUserInformation/HeaderUserInformation.tsx
@@ -75,7 +75,7 @@ const HeaderUserInformation: FC<HeaderUserInformationProps> = ({
       id={id}
       role={role}
       aria-haspopup={popUpType}
-      {...(popUpId ? { 'aria-controls': popUpId } : {})}
+      aria-controls={popUpId}
       aria-expanded={isPopUpOpen}
     >
       <Tooltip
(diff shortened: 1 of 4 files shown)


src/HeaderUserInformation/HeaderUserInformation.tsx
```diff
@@ -36,8 +36,21 @@ export interface HeaderUserInformationProps {
...
+  popUpType?:
+    | boolean
+    | 'dialog'
+    | 'menu'
+    | 'grid'
+    | 'listbox'
+    | 'false'
+    | 'true'
+    | 'tree'
+    | undefined;
```
dialexo, line 51: Not sure why the popup type is manually enumerated here? Is this derived from an existing type? Also you don't need to specify `| undefined` if the prop is marked as optional with the `?` annotation.

dialexo (dismissed):

pyphilia (approved):
LGTM 👾

src/HeaderUserInformation/HeaderUserInformation.tsx
```diff
@@ -62,7 +70,14 @@ const HeaderUserInformation: FC<HeaderUserInformationProps> = ({
   }
 
   return (
-    <WrapperDiv onClick={onClick} id={id}>
+    <WrapperDiv
+      onClick={onClick}
+      id={id}
+      role={role}
+      aria-haspopup={popUpType}
+      {...(popUpId ? { 'aria-controls': popUpId } : {})}
```
spaenleh: You can simplify it to: `aria-controls={popUpId}` since if the popUpId is undefined it is equivalent for the prop.

src/utils/fixtures.ts
```diff
@@ -9,3 +10,13 @@ export const MOCK_MEMBER = {
...
+
+export const MOCK_MEMBER_RECORD = {
+  id: 'id',
+  name: 'creator',
+  email: 'email',
+  extra: {},
+  createdAt: new Date(),
+  updatedAt: new Date(),
+  type: MemberType.Individual,
+} as MemberRecord;
```
spaenleh: You should not cast an object to a MemberRecord. You should use the `convertJs` function from sdk to convert from a JS object to an immutable and set the type of the variable to `MemberRecord`.

spaenleh (approved):
👓 Looks good, see my comments, I fixed them already, so it is good to merge !

PR #415, opened 2023-06-30
fix(a11y): use item name when alttext is falsy

spaenleh:
The `altText` should default to the item name when it is `undefined` or `''`. This is not the best but at least assistive technologies will not skip it.


MalinSvenberg (approved):
LGTM!

PR #460, opened 2023-07-27
fix: zIndex on quill tooltip

spaenleh:
In this PR we fix an issue related to the `z-index` property on the elements in the Quill Text Editor.

### Issue:
When opening the link tool, the popup that appears is **under** the "Cancel" and "Save" buttons. 

### Solution:
Set a `z-index` higher than the z-index of the buttons. We can use [the z-index from MUI](https://mui.com/material-ui/customization/z-index/) and select a suitable value like `tooltip`. 

Screenshot 2023-07-27 at 19 09 55

### DIFF
--- a/src/TextEditor/TextEditor.tsx
+++ b/src/TextEditor/TextEditor.tsx
@@ -64,38 +64,35 @@ export type TextEditorProps = {
   value?: string;
 };
 
-const Div = styled('div')(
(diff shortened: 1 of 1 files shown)


swouf (approved):
Cool 😎 Thanks for fixing this so quickly!

PR #464, opened 2023-07-28
chore(ci): add prettier check and rename steps

spaenleh:


.github/workflows/test.yml
```diff
@@ -7,22 +7,33 @@ jobs:
...
+
+      - name: Check code style and linting
+        run: yarn prettier:check && yarn lint
+
+      - name: Build # ensures type-checks and compiles
         run: yarn build
-      - name: install playwright
+
+      - name: Install playwright
         run: npx playwright install --with-deps
```
swouf, line 31: Just out of curiosity, why do we install *playwright* here and at this point?

swouf (approved):
Well done 👍🏼 The workflows are definitely cleaner now. Thanks.

.github/workflows/test.yml
spaenleh: Playwright is used to perform storybook "interaction tests" (see [this doc](https://storybook.js.org/docs/react/writing-tests/interaction-testing)). It allows you to interact with your component. As it is only a development/testing library we do not need to install it for building the source code. Also it can be large as a dependency.

PR #520, opened 2023-09-20
feat: table pagination

### LINKED ISSUE #519: Table component Allow Backend Pagination

LinaYahya:
Add total count as a props for table component and keep items length as a default value, as well as control pagination functionality from external props
closes #519


src/TableComponent/Table.tsx
```diff
@@ -276,15 +284,13 @@ function GraaspTable<T>({
         {/* TODO: implement pagination */}
         {pagination && (
           <TablePagination
-            count={rowData.length}
+            count={totalCount ? totalCount : rowData.length}
```
pyphilia, line 287:
```suggestion
            count={totalCount ?? rowData.length}
```

src/TableComponent/Table.tsx
```diff
@@ -276,15 +284,13 @@ function GraaspTable<T>({
...
             labelDisplayedRows={labelDisplayedRows}
-            page={0}
+            page={isClientPagination ? 0 : page}
+            onPageChange={onPageChange ? onPageChange : () => {}}
```
pyphilia, line 290:
```suggestion
            onPageChange={onPageChange ?? () => {}}
```

src/TableComponent/Table.tsx
```diff
@@ -276,15 +284,13 @@ function GraaspTable<T>({
...
             rowsPerPageOptions={[]}
-            onPageChange={() => {
-              // todo
-            }}
             component='div'
-            rowsPerPage={-1}
+            rowsPerPage={isClientPagination ? -1 : 10}
```
pyphilia, line 293: `10` comes from your default limit right? I think it should be a prop as well.
Also why not use https://github.com/graasp/graasp-sdk to have the default value (to be reused in the backend as well)?

pyphilia (changes requested):
Thanks for the PR!

pyphilia (changes requested):
Thanks for the PR! I think it needs a subtle change 😄

PR #533, opened 2023-09-27
fix: improve getUUID and add tests

spaenleh:
- fix: first test that the uuid is a long form
- fix: add tests and re-enable jest unit test in workflow and locally

Example of an error message in the console in Builder when visiting a page with a long UUID. It can be miss-leading as it is not really an error.
Screenshot 2023-09-27 at 15 59 54


pyphilia (approved):
LGTM

PR #549, opened 2023-10-12
feat: add onclick for linkItem component

LinaYahya:
Pass onClick to link item as a props, As we might do specific action on clicking item link (basically analytics triggers)


spaenleh (changes requested):
We should be careful when using onClick with links. In this case it will work since we target of the link is `_blank` so it opens a new tab and the JS has time to execute.
In cases where the JS performs an async operation and the link loads a new page before the end of the operation it can be more complicated.

spaenleh (changes requested):
Change look good, can we apply the same to the other places ?

src/items/DownloadButtonFileItem.tsx
```diff
@@ -14,13 +14,15 @@ interface DownloadButtonFileItemProps {
   url?: string;
   id?: string;
   text?: string;
+  handleLoad?: () => void;
```
spaenleh, line 17: Can we change this prop name to `onClick` ? or was there a particular reason why it is called `handleLoad` ?

src/items/FileItem.tsx
```diff
@@ -53,6 +53,7 @@ export interface FileItemProps {
   showCaption?: boolean;
   showCollapse?: boolean;
   sx?: SxProps;
+  handleLoad?: () => void;
```
spaenleh, line 56: Same here

spaenleh (changes requested):

spaenleh (approved):

PR #550, opened 2023-10-19
feat: add custom profile option to user dropdown

LinaYahya:
Update `UserSwitchWrapper` to render both liked Items and library profile options conditionally.

### DIFF
--- a/src/UserSwitch/UserSwitch.tsx
+++ b/src/UserSwitch/UserSwitch.tsx
@@ -20,7 +20,6 @@ import React, {
 import { isPseudonymizedMember } from '@graasp/sdk';
 import { MemberRecord } from '@graasp/sdk/frontend';
 
-import Button from '../buttons/Button';
 import { SHORT_TEXT_WIDTH, SMALL_AVATAR_SIZE } from '../constants';
 import { Variant } from '../types';
 
@@ -44,10 +43,7 @@ interface Props {
   members?: List<MemberRecord>;
   menuId?: string;
   onMemberClick?: (_id: string) => MouseEventHandler;
-  onSeeProfileClick?: MouseEventHandler;
   renderAvatar?: (member?: MemberRecord) => JSX.Element;
-  seeProfileButtonId?: string;
-  seeProfileText?: string;
   signedOutTooltipText?: string;
 }
 
@@ -58,10 +54,7 @@ const UserSwitch: FC<Props> = ({
   isMemberLoading = false,
   member,
   menuId,
-  onSeeProfileClick,
   renderAvatar = () => <></>,
-  seeProfileButtonId,
-  seeProfileText = 'See Profile',
   signedOutTooltipText = 'You are not signed in.',
 }) => {
   const [anchorEl, setAnchorEl] = useState<(EventTarget & Element) | null>(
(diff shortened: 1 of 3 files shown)


src/UserSwitch/UserSwitchWrapper.tsx
spaenleh, line 1: I would prefer to have a prop called `menuItems` that would be an array of type:
```ts
type UserMenuItem = {
  icon: JSX.Element;
  text: string;
  redirect_path: string;
  // ... other properties like `id` to make it easily selectable in tests
}
```

This way builder, player etc can input whatever custom menu items they want.
This will remove duplication and allow more flexibility.
This will also mean you can remove the `null` from the array as it is not really necessary.

src/UserSwitch/UserSwitch.tsx
```diff
@@ -35,7 +34,7 @@ const StyledWrapper = styled(Box)(() => ({
 }));
 
 interface Props {
-  Actions?: JSX.Element | JSX.Element[];
+  Actions?: JSX.Element | (JSX.Element | null)[];
```
spaenleh, line 37: I don't think there is a need for this change. The array should not contain `null` values.

src/UserSwitch/UserSwitchWrapper.tsx
```diff
@@ -96,17 +113,61 @@ const UserSwitchWrapper: FC<Props> = ({
...
+          key='libraryProfile'
+          onClick={goToLibraryProfile}
+          id={signOutMenuItemId}
+        >
+          <ListItemIcon>
+            <ProfileIcon fontSize='large' />
+          </ListItemIcon>
+          <Typography variant='subtitle2'>{libraryProfileText}</Typography>
+        </MenuItem>
+      ) : null,
```
spaenleh, line 158: I don't think we should have `null`s in the array, instead can we just push the values conditionally ?

spaenleh (changes requested):
✉️ Thank you for these changes, I would suggest a change in the interface to allow more menues to be added in the future.

src/UserSwitch/UserSwitchWrapper.tsx
LinaYahya: @spaenleh 
I did this as I want to have it custom places, middle as an example, and also null is a valid jsx element

spaenleh: @LinaYahya You could choose the place with the order in the `menuItems` array prop: 
```ts
<UserSwitchWrapper
  menuItems={
    [
      { icon: <ProfileIcon />, text: 'PublicProfile', redirection_path: '/members/1234567' },
      { icon: <LikeIcon />, text: 'Liked Items', redirection_path: '/liked' },
      // ... etc
    ]
  }
  // other props
/>

spaenleh: It is true `null` is a valid JSX.Element but does it really make sens to populate an array with it when you know there will be no element there ?

spaenleh (commented):
Thank you for the changes, the structure and the props look good to me !

Just not sure we want to use the "Gear/Settings" icon for the "See Profile" menu
Screenshot 2023-10-23 at 14 07 10

 Personally I would rather use the [Account Circle](https://mui.com/material-ui/material-icons/?query=profile&selected=AccountCircle), wdyt @pyphilia ?

spaenleh (approved):
LGTM ! 🎉

PR #628, opened 2024-01-03
style: Responsive enhancements

### LINKED ISSUE #629: Switch platform component | responsive enhancements
- enhance platform switch component at mobile view to use [MUI Speed dial component](https://mui.com/material-ui/react-speed-dial/) for mobile view
- create a hook to detect mobile view

LinaYahya:
- enhance platform switch component at mobile view to use [MUI Speed dial component](https://mui.com/material-ui/react-speed-dial/) for mobile view

closes #629


src/hooks/useMobileView.tsx
spaenleh, line 1: I am not sure we need to have this hook. Could we get away with the [`useMediaQueryHook`](https://mui.com/material-ui/react-use-media-query/#using-material-uis-breakpoint-helpers) provided by MUI ?

src/UserSwitch/UserSwitch.tsx
```diff
@@ -121,9 +123,11 @@ const UserSwitch: FC<Props> = ({
...
 
         <div>
-          <Typography variant='h6' noWrap>
-            {memberName}
-          </Typography>
+          {!isMobile && (
+            <Typography variant='h6' noWrap>
+              {memberName}
+            </Typography>
+          )}
```
spaenleh, line 130: This name display is the one displayed inside the Popover. If you were trying to not show the name of the user next to the avatar, this is not the right place. You might be looking for line 170. See the screenshot below
Screenshot 2024-01-05 at 15 26 55

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -79,6 +80,7 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
   platformsProps,
 }) => {
   /** Helper inner component: generates buttons from icons while capturing parent props */
+  const { isMobile } = useMobileView();
```
spaenleh, line 83: Move this line up, so the comment that refers to the component below stays relevant.

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -157,6 +158,57 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
     />
   ));
 
+  if (isMobile) {
+    const SelectedIcon = PlatformIcons[selected || 'Builder'];
```
spaenleh, line 162:
```suggestion
    const SelectedIcon = PlatformIcons[selected || Platform.Builder];
```

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -157,6 +158,57 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
...
+        >
+          {Object.values(Platform).map((platform, index) => {
+            const Icon = PlatformIcons[platform];
+            const isSelectedPlatform = platform === selected;
+            const platformProps = platformsProps?.[platform];
+            const iconProps = {
+              size,
+              secondaryColor: isSelectedPlatform ? color : accentColor,
+              primaryColor: isSelectedPlatform ? accentColor : color,
+              fill: isSelectedPlatform ? color : accentColor,
```
spaenleh, line 185: I would prefer to have the colors inverted, so the selected platform is on a blue background and the non selected platforms are on a white background:
Screenshot 2024-01-05 at 15 36 04

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -157,6 +158,57 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
...
+    return (
+      <Box sx={{ position: 'relative' }}>
+        <SpeedDial
+          ariaLabel='SpeedDial basic example'
+          sx={{
+            position: 'absolute',
+            '& button': {
+              width: '40px',
+              height: '40px',
+            },
```
spaenleh, line 172: You can remove this part if you pass the `size="small"` prop to the FabProps prop. It renders the Floating Action button (the component that powers the main speed dial button as a "dense" button.

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -157,6 +158,57 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
...
+          }}
+          icon={selected && <SelectedIcon />}
```
spaenleh, line 174: In order to make the speed dial more visible on the blue banner you can add a white border to the speed dial button. This has the added advantage of matching the current platform design which has a border.
```suggestion
          }}
          FabProps={{ size: 'small', sx: { border: '2px solid white' } }}
          icon={selected && <SelectedIcon />}
```

src/PlatformSwitch/PlatformSwitch.tsx
spaenleh: Actually I got confused by the fact that the "right" color scheme is in the "dark" story. But when viewing it in "light" mode it looks right in the expanded version but not the mobile one... So we should either remove the light version story, or use a different background color to match the effect that the expanded version has. Not sure this is clear ...

spaenleh (changes requested):
Nice work, 💪  see my comments.

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -159,20 +168,19 @@ export const PlatformSwitch: FC<PlatformSwitchProps> = ({
   ));
 
   if (isMobile) {
-    const SelectedIcon = PlatformIcons[selected || 'Builder'];
+    const SelectedIcon = PlatformIcons[selected || Platform.Builder];
     return (
-      <Box sx={{ position: 'relative' }}>
+      <Box sx={{ position: 'relative', height: '40px' }}>
```
spaenleh, line 173: I am not sure why this is needed, maybe you can get rid of the box completely and add the `sx={{ position: "relative" }}` to the SpeedDial directly ?

spaenleh (changes requested):
🥣 Nice ! Just a small comment still.

src/PlatformSwitch/PlatformSwitch.tsx
spaenleh: @LinaYahya I see that without it the header takes the height of the speeddial, I think it is ok to keep it. maybe try to remove the box and see if adding the relative position on the speeddial directly can fix it, otherwise this is fine.

spaenleh (approved):
See https://github.com/graasp/graasp-ui/pull/628#discussion_r1444502603 I approve now, so you can merge after you do the test to see if you can remove the hardcoded height.

src/PlatformSwitch/PlatformSwitch.tsx
LinaYahya: Unfortunately, it didn't work as the speed dial has an absolut position, so it should be wrapped within the relative container

PR #631, opened 2024-01-05
feat: add disable to select options

### LINKED ISSUE #630: Modify Select Component to have disabled at values props
Modify `select` component to have `disabled` at values, which will be optional property

LinaYahya:
closes #630

### DIFF
--- a/src/Select/Select.tsx
+++ b/src/Select/Select.tsx
@@ -16,7 +16,7 @@ type Props<T> = {
   labelId?: MuiSelectProps['labelId'];
   onChange?: MuiSelectProps<T>['onChange'];
   sx?: SxProps;
-  values: { value: T; text: string }[];
+  values: { value: T; text: string; disabled?: boolean }[];
   variant?: MuiSelectProps['variant'];
   size?: MuiSelectProps['size'];
 };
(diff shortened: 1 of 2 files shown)


src/Select/Select.tsx
```diff
@@ -57,11 +57,12 @@ function Select<T extends string | number | readonly string[] | undefined>({
         {...(value && { value })}
         sx={sx}
       >
-        {values.map(({ value, text }) => (
+        {values.map(({ value, text, disabled }) => (
           <MenuItem
             key={buildOptionId?.(value)}
             id={buildOptionId?.(value)}
             value={value}
+            disabled={Boolean(disabled)}
```
spaenleh, line 65: Actually I don't think you need to Cast it to a Boolean as the `boolean | undefined` type of the argument already matches the behaviour of the `disabled` prop. Feel free to keep it however, if you think it makes it more clear.

spaenleh (changes requested):
Thank you for this change 💮 

It would be nice when doing such changes to also update the story to display the possible options and configurations that components can have. I have pushed some changes on the [`630-select-disabled-story`](https://github.com/graasp/graasp-ui/pull/632) branch for you to see, I think it would be a nice addition. Maybe you could add a new story to display the disabled option ?

spaenleh (approved):

PR #640, opened 2024-01-10
fix: disable select if all options are disabled

LinaYahya:


spaenleh:
Oh looks like our PRs got crossed. I added a disabled prop in https://github.com/graasp/graasp-ui/pull/634 and in the Builder I allowed the `disabled` prop to be either a `boolean` or the `DisabledMap` that you created and passed it to the right Select prop. 

Here you propose something that makes the Select component a bit smarter.

Should we keep both @pyphilia ?

spaenleh (changes requested):
Changes look good, but can you rebase on the latest main ?

spaenleh (approved):

PR #644, opened 2024-01-11
feat: add mimetype option in `ItemIcon`

spaenleh:
This `mimetype` option is an alternative to providing the `extra` prop


pyphilia (approved):
For later optimization of UI, you'll have to remove these `window` I think.

PR #665, opened 2024-01-23
feat: update `Main` component

### LINKED ISSUE #636: Refactor Main Component
The Main component is not very flexible and is still a class component (woo very old stuff).
- [ ] change usage of `Grid` to `Stack`
- [ ] allow elements to use ellipsis and never overflow on a second line

spaenleh:
In this PR:
- fix #636 from a class to a functional component
- add support for mobile `temporary` drawer for screen smaller (above `sm` breakpoint)
  - the drawer will open "on top" of the content with a backdrop and closes once an option has been selected, or the user clicks out or the user closes the drawer with the button
- add support for a closable drawer that "pushes" content to the side when opened on large screens
- add support for ellipses in the title when space is constrained
- add animations to the width changes
- add a story to show the Desktop and Mobile views with come content

Demo video bellow:

https://github.com/graasp/graasp-ui/assets/39373170/61451517-f6b0-4878-a3c1-cb880899e5e0


src/Main/Main.tsx
```diff
@@ -1,141 +1,177 @@
...
+          <Stack
+            width='100%'
+            direction='row'
+            alignItems='center'
+            justifyContent='space-between'
+            spacing={2}
+            flex={1}
+            minWidth='0px'
+          >
+            <Stack direction='row' alignItems='center' flex={1} minWidth='0px'>
```
pyphilia, line 119:
```suggestion
            <Stack direction='row' alignItems='center' flex={1} minWidth={0}>
```

src/Main/Main.tsx
```diff
@@ -1,141 +1,177 @@
...
+      >
+        <Toolbar>
+          <Stack
+            width='100%'
+            direction='row'
+            alignItems='center'
+            justifyContent='space-between'
+            spacing={2}
+            flex={1}
+            minWidth='0px'
```
pyphilia, line 117:
```suggestion
            minWidth={0}
```

src/Main/Main.tsx
```diff
@@ -1,141 +1,177 @@
...
+          >
+            <Stack direction='row' alignItems='center' flex={1} minWidth='0px'>
+              <IconButton
+                color='inherit'
+                aria-label='open drawer'
```
pyphilia, line 122: Translate?

src/Main/Main.tsx
```diff
@@ -1,141 +1,177 @@
...
+          </Stack>
+        </Toolbar>
+      </AppBar>
+      <Box component='nav' aria-label='navigation'>
+        <Drawer
+          variant='temporary'
+          open={open}
+          onClose={handleDrawerClose}
+          ModalProps={{
+            keepMounted: true, // Better open performance on mobile.
```
pyphilia, line 141: Does it make sense to have it false on desktop then? Just asking.

src/Main/Main.tsx
```diff
@@ -1,141 +1,177 @@
...
+              >
+                <MenuIcon />
+              </IconButton>
+              {headerLeftContent}
+            </Stack>
+            {headerRightContent}
+          </Stack>
+        </Toolbar>
+      </AppBar>
+      <Box component='nav' aria-label='navigation'>
```
pyphilia, line 135: Translate.

pyphilia (changes requested):

src/Main/Main.tsx
```diff
@@ -1,141 +1,177 @@
...
+const drawerWidth = 240;
 
-const DrawerHeaderContainer = styled('div')(({ theme }) => ({
-  display: 'flex',
-  alignItems: 'center',
-  padding: '0 8px',
-  ...theme.mixins.toolbar,
-  justifyContent: 'flex-end',
+const buildHeaderGradient = (color: string): string =>
+  `linear-gradient(90deg, ${PRIMARY_COLOR} 0%, ${PRIMARY_COLOR} 35%, ${color} 100%);`;
```
ReidyT, line 19: It is wanted to have primary color twice (0% and 35%) ? If yes, shouldn't be possible to only use `linear-gradient(90deg, ${PRIMARY_COLOR} 35%, ${color} 100%);` ?

ReidyT (approved):
Amazing work 🏢 !

src/Main/Main.tsx
spaenleh: Actually this just means that we use the `keepMounted` option of the dialog. So it will not be destroyed when closed. This is to improve performance on mobile. It also improves performance on desktop. But in this case this specific variant of the drawer is not shown on large screens (display: none).

src/Main/Main.tsx
spaenleh: I think the idea was to have a gradient that would not be symmetrical, more blue on the left side. I tried and it looks like I can get rid of the '0%' color indication. So maybe that is what I will do ! Thank you for the comment.

src/Main/Main.tsx
```diff
@@ -1,141 +1,217 @@
...
+    color='inherit'
+  >
+    <GraaspLogo height={40} sx={{ fill: 'white' }} />
+    <Typography sx={{ display: { xs: 'none', sm: 'block' } }} variant='h6'>
+      Graasp
+    </Typography>
+  </Stack>
+);
+
+const drawerWidth = 240;
```
pyphilia, line 32:
```suggestion
const DRAWER_WIDTH = 240;
```

src/Main/Main.tsx
```diff
@@ -1,141 +1,217 @@
...
-const DrawerHeaderContainer = styled('div')(({ theme }) => ({
-  display: 'flex',
-  alignItems: 'center',
-  padding: '0 8px',
-  ...theme.mixins.toolbar,
-  justifyContent: 'flex-end',
+import GraaspLogo from '../GraaspLogo';
+import { AccentColors, PRIMARY_COLOR } from '../theme';
+
+const LogoHeader = (): JSX.Element => (
```
pyphilia, line 18: Should this be in a separate file (still within /Main)?

pyphilia (approved):

PR #669, opened 2024-01-23
fix!: use `domLayout='autoHeight'` to autosize the table

### LINKED ISSUE #668: Allow AG grid table to auto-height
https://www.ag-grid.com/javascript-data-grid/grid-size/#grid-auto-height

spaenleh:
In this PR:
- fix #668 

And I remove the `tableHeight` prop of the Table component since it is not necessary anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "post-commit": "git status",
     "pre-commit": "yarn check",
     "pre-push": "yarn build",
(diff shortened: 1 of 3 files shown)


spaenleh:
This "improvement" caused the `ResizeObserver loop completed with undelivered notifications` error in Cypress tests. 

Since we are working on using another library for the table, we will close this.

PR #684, opened 2024-01-31
fix: update forbiddenContent

spaenleh:
According to a comment in builder I remove some props here. I checked usage and it looks like the `showPseudonymized` option is not used in Builder nor player.


ReidyT (approved):
Nice refactor ! 🚀

pyphilia (approved):

PR #685, opened 2024-02-01
fix: open and dismiss menu and add relative position to content wrapper

spaenleh:
In this PR:
- add the `display: flex` property on the toplevel element of the `Main` component to allow flex grow to work in the children content.
- Split the component to use a new context that manages the open state so it is possible for `MenuItem` to dismiss the drawer when clicking on an element in mobile view
- use `position: relative` to allow the Uppy container to be placed in absolute position.


ReidyT (approved):
Looks good 🙌

PR #688, opened 2024-02-02
feat: navigator render extra item

### LINKED ISSUE #686: Add item settings menu inside navigator

LinaYahya:
Add an extra props for navigation component to accept extraItems, Basically `extraItems` has the following structure
```
[
  {
    name: "",
    path?: "",
    Icon?: ,
    menuItems?: [
      {
        name: "",
        path: ""
      }
    ]
  }
]

```
image

closes #686

### DIFF
--- a/src/Navigation/Navigation.tsx
+++ b/src/Navigation/Navigation.tsx
@@ -1,6 +1,6 @@
 import truncate from 'lodash.truncate';
 
-import { SvgIconTypeMap, SxProps } from '@mui/material';
+import { Box, SvgIconTypeMap, SxProps } from '@mui/material';
 import Breadcrumbs from '@mui/material/Breadcrumbs';
 import Typography from '@mui/material/Typography';
 import { styled } from '@mui/material/styles';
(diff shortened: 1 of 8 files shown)


src/Navigation/Navigation.tsx
```diff
@@ -1,12 +1,14 @@
...
-import { SxProps } from '@mui/material';
+import { Box, SvgIconTypeMap, SxProps } from '@mui/material';
 import Breadcrumbs from '@mui/material/Breadcrumbs';
 import Typography from '@mui/material/Typography';
 import { styled } from '@mui/material/styles';
+import { OverridableComponent } from '@mui/types';
 
 import { DiscriminatedItem, ItemType } from '@graasp/sdk';
 
+import ItemActionsMenu from './ExtraItemsMenu';
```
spaenleh, line 11: Why does this not have the same name as the file ?

src/Navigation/Navigation.tsx
```diff
@@ -30,8 +32,18 @@ export type NavigationProps = {
...
+  extraItems?: ExtraItem[];
 };
-
+export interface MenuItemType {
+  name: string;
+  path: string;
+}
+export interface ExtraItem {
+  name: string;
+  path?: string;
```
spaenleh, line 43: Why is this optional ?

spaenleh (changes requested):
Overall I think your implementation is ok and it does the job. 

But as I am looking at the code for this component, It does not look very flexible, the code is very convoluted et we are fighting MUI defaults so hard... 

I also expressed by concerns about the fact that I do not think this component is intuitive (I never would have guessed that I can click on the "separator" to navigate to a children...

In my opinion we should re-architect it to
1. improve the component API
2. improve the look and usability of the component
3. allow extensibility and composition in the future

I will not ask you to do that in this PR obviously but this is something we should discuss together with @pyphilia and @ReidyT as well as Martina.

PR #689, opened 2024-02-02
fix: use sx prop passed when there is an src

spaenleh:
In this PR:
- use the `sx` prop passe when displaying the `iconSrc`
- remove deprecated props

### DIFF
--- a/src/icons/ItemIcon.stories.tsx
+++ b/src/icons/ItemIcon.stories.tsx
@@ -1,6 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/react';
 
-import { ItemType, MimeTypes, S3FileItemExtra } from '@graasp/sdk';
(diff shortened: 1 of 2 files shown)


src/icons/ItemIcon.stories.tsx
```diff
@@ -40,63 +44,94 @@ export const Default: Story = {
...
+    type: ItemType.S3_FILE,
+    color: 'red',
+    sx: { fontSize: '3rem' },
     extra: {
       [ItemType.S3_FILE]: {
-        mimetype: MimeTypes.ZIP,
+        name: '',
+        mimetype: MimeTypes.Image.JPEG,
       },
     } as S3FileItemExtra,
```
pyphilia, line 136: Is it to be sure it works with legacy code? Maybe add a comment that we can remove this test later.

pyphilia (approved):

PR #691, opened 2024-02-05
fix: use `overflow: auto` instead of `overflow: scroll`

### LINKED ISSUE #690: Use `overflow: auto` to not show scrollbars when not neeeded
This is a visual "bug" that is related to the page leaving space for the scroll bars.

If the os (macOS) is set to "Show scroll bars: always" **and** the site sets `overflow: scroll` on an element, the browser will resever some space for the scollbars even when no scrolling is necessary.

This behavior is not visible when the OS setting is "when scrollling" since it overlays the scrollbars on top of the content as they disapear when not in use.

We should use the `overflow: auto` property to not show the padding when no scollbars are necessary.

spaenleh:
In this PR:
- fix #690


pyphilia (approved):

ReidyT (approved):
👍

Issue #735, opened 2024-03-04
FileItem image changes size depending on showCaption


ReidyT:
https://github.com/graasp/graasp-ui/assets/147397675/57029ac4-2f8b-48f9-a072-3455c5b9d5c1

PR #739, opened 2024-03-04
fix: wrap component inside Box in withCaption to solve image height

ReidyT:
- Should solve issue #735.

## Before
image

## After
image


spaenleh (approved):
Thank you !

PR #756, opened 2024-03-13
feat(apps): error fallback form

swouf:


src/appComponents/ErrorFallback.tsx
```diff
@@ -0,0 +1,193 @@
...
+  nameLabel = NAME_LABEL,
+  nameHelper = NAME_HELPER,
+  emailLabel = EMAIL_LABEL,
+  emailHelper = EMAIL_HELPER,
+  commentLabel = COMMENT_LABEL,
+  commentHelper = COMMENT_HELPER,
+  thanksMessage = THANKS_FOR_FEEDBACK,
+  sendButtonLabel = SEND,
+  errorDetailsLabel = ERROR_DETAILS,
+}) => {
```
spaenleh, line 108: We prefer the style that does not use the `FC`, I know we switched a while back and you used to declare them like that, and you were right :)
```suggestion
const ErrorFallback = ({
  error,
  componentStack,
  eventId,
  captureUserFeedback,
  title = TITLE,
  formTitle = FORM_TITLE,
  nameLabel = NAME_LABEL,
  nameHelper = NAME_HELPER,
  emailLabel = EMAIL_LABEL,
  emailHelper = EMAIL_HELPER,
  commentLabel = COMMENT_LABEL,
  commentHelper = COMMENT_HELPER,
  thanksMessage = THANKS_FOR_FEEDBACK,
  sendButtonLabel = SEND,
  errorDetailsLabel = ERROR_DETAILS,
}: ErrorFallbackProps): JSX.Element => {
```

src/appComponents/ErrorFallback.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+const meta: Meta<typeof ErrorFallback> = {
+  title: 'Apps/ErrorFallback',
+  component: ErrorFallback,
+};
+
+export default meta;
+
+type Story = StoryObj<typeof ErrorFallback>;
+
+export const DefaultErrorFallback: Story = {
```
spaenleh, line 15: Add a mobile view story ? you can set the viewport: 
```ts
args: {
...
},
parameters: {
    viewport: {
      defaultViewport: 'mobile2',
    },
  },
```

src/appComponents/ErrorFallback.tsx
```diff
@@ -0,0 +1,193 @@
...
+              fontSize='1.8rem'
+            >
+              {title}
+            </Typography>
+          </Stack>
+          {feedbackGiven ? (
+            <Alert severity='success'>{thanksMessage}</Alert>
+          ) : (
+            <Box id='user-feedback'>
+              <Stack direction='column' spacing={1} maxWidth='82rem'>
```
spaenleh, line 144: Where does that max-width come from ? Seems a bit arbitrary no ?

spaenleh (approved):
Looks good, just some nitpicks, thank you for this improvement !

src/appComponents/ErrorFallback.tsx
swouf: That's true, it's completely arbitrary. It was because on very large screen, the form looks way too wide. Do you have any suggestion?

PR #805, opened 2024-04-15
feat: expose theme options for next

spaenleh:
In this PR I allow to set the font-family property on the theme. 

The project that use Next need to be able to set them as they use different names for the fonts. For more details see https://github.com/graasp/graasp-library/pull/594 

I also added the fontfile so it is possible to use the project without a connection to the google CDN Servers


pyphilia (approved):
🦆

PR #814, opened 2024-04-19
fix: propose general improvements related to player

spaenleh:
In this PR:
- add `backgroundColor` option for main and drawers
- update usages of name to be displayName where relevant (this is a bit of an issue in some places, comments welcome)
- default to button text using `capitalise` instead of `uppercase`

### DIFF
--- a/src/Collapse/withCollapse.tsx
+++ b/src/Collapse/withCollapse.tsx
@@ -1,8 +1,14 @@
 import Collapse from './Collapse';
 
-const withCollapse = <T extends { name: string }>({ item }: { item: T }) => {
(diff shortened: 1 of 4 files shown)


spaenleh:
@pyphilia the strategy will be to have display name for all types of item. There are 2 PRs open in Builder to add these changes for folder and link. We need to implement a new route for create folder with thumbnail (using formData) in the backend in order to merge these PRs. So what I did in this code is to use display name where we have components in the player, and when the components were used in both i kept name and will provide name with the value of display name instead. This might be a bit confusing, not sure how to make it better.

PR #824, opened 2024-04-24
feat: add row menu, row menus and breadcrumbs

pyphilia:
Factor out of builder's item selection modal

### DIFF
--- a/src/Tree/RowMenu.tsx
+++ b/src/Tree/RowMenu.tsx
@@ -1,0 +1,110 @@
+import { KeyboardArrowRight } from '@mui/icons-material';
+import {
+  Button,
+  CircularProgress,
+  IconButton,
+  Stack,
+  Typography,
+  styled,
+} from '@mui/material';
+
+import { useState } from 'react';
+
+import { ItemType } from '@graasp/sdk';
+import { ItemIcon } from '@graasp/ui';
+
+import type { NavigationElement } from './types';
+
+export interface RowMenuProps {
+  item: NavigationElement;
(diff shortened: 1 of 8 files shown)


src/Tree/RowMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+const RowMenu = ({
+  item,
+  onNavigate,
+  onClick,
+  selectedId,
+  isDisabled,
+  id,
+  arrowId,
+}: RowMenuProps): JSX.Element | null => {
+  const [isHoverActive, setIsHoverActive] = useState(false);
```
spaenleh, line 48: I think a state is not needed. See this demo, you can add a bit of css to toggle the visibility of the IconButton when the parent is hovered. I should be more responsive (less latency) and reduce code and logic.

https://github.com/graasp/graasp-ui/assets/39373170/c7bb0359-64b6-4ef7-bb38-79cbbf6950e0

src/Tree/types.ts
```diff
@@ -0,0 +1,7 @@
+export type NavigationElement = {
+  id: string;
+  name: string;
+  // important to have to compute if an element should be disabled
+  path: string;
```
spaenleh, line 5: You already pass a isDisabled function no ? can this be removed ?

spaenleh (approved):
Looks good, I think we can improve a few things, otherwise very nice !

src/Tree/types.ts
pyphilia: Yes, but it's the data I send in the arguments of the function, so I need it ;)

PR #825, opened 2024-04-25
fix: remove margin when there is no label

spaenleh:
In this PR I remove the margin that is present when there is no label.


ReidyT (approved):
Nice ! LGTM 👍

PR #834, opened 2024-04-29
fix: playwright install issue

spaenleh:
It looks like when playwright is added from the cache the test complain because there is something missing. So in this PR I install playwright both if the cache is found and if it is missing. Hopefully this fixes the issue.


ReidyT (approved):
LGTM 🚀

PR #843, opened 2024-05-02
feat: update deps

pyphilia:
Updated some deps and fix accordingly.


spaenleh (approved):
So we are now on storkybook 8 ? 🥳

PR #872, opened 2024-05-24
feat(apps): add status chips

swouf:


ReidyT (approved):
Nice PR ! It looks good to me 🚀. I just have a question: do you think it could be interesting to have a StatusChip component who have the same props as the others, but with a colour and an icon in addition ? This could allows to reuse this component inside Required, Saved and SubmittedChip component and reducing the little code duplication. Maybe it is not what we want, it is just a question 😉.

spaenleh (commented):
@ReidyT @swouf I agree, we could have a single component that exposes a `variant` prop that could take the `required | submitted | saved` values and then simply switch on them in the component. The structure could be the same and you only have to switch on the color and icon properties. You can put them in a dictionary that indexes over the custom chip variant. 

But as Thibault has said, do this only if re-use is important.

swouf:
@ReidyT @spaenleh, thanks for the recommendations. I agree with your suggestion but I wonder if we will add features that differ a lot for each type of chip. As I don't know that currently, I will merge this PR as is but turn your comments into an issue. Thanks again.

PR #884, opened 2024-05-31
fix: allow fullscreen for h5p iframe

pyphilia:
Some h5p provide fullscreen feature. Before the fix it resized once and stays like this, it's impossible to revert.

https://github.com/graasp/graasp-ui/assets/11229627/eda5ac7f-83ea-48c9-afc5-05a0e5188a34

I added the `allowFullscreen` prop for the iframe.
It does not work perfectly, one click over three resize weirdly. 

Without this change, the fullscreen does not work at all, so I guess it's okay as compromise.
I quickly tried to set `height=100%` on fullscreen event, but it didn't work.

https://github.com/graasp/graasp-ui/assets/11229627/09f9ebba-74be-4759-80d8-acf84dd31ff5

### DIFF
--- a/src/items/H5PItem.tsx
+++ b/src/items/H5PItem.tsx
@@ -97,9 +97,13 @@ const H5PItem: FC<H5PItemProps> = ({
       ref={iframeRef}
       id={iframeId}
       src={integrationUrl.href}
-      scrolling={'no'}
-      frameBorder={0}
-      style={{ width: '100%', border: 'none', display: 'block' }}
+      allowFullScreen
+      style={{
+        width: '100%',
+        border: 'none',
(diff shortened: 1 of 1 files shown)


ReidyT (commented):
I just have a question about the last video. With this change, the user can easily enter in fullscreen, but then he has to click twice to exit this mode ?

spaenleh (approved):
Looks good ! I had a user show me this behaviour too, nice that you were able to fix it.

pyphilia:
> I just have a question about the last video. With this change, the user can easily enter in fullscreen, but then he has to click twice to exit this mode ?

yes that's the only pb =/

PR #886, opened 2024-06-03
fix: unset height when using dynamic resizing

### LINKED ISSUE #885: [BUG] App iframe does not resize
Due to a change made in #854 to the common Styled iframe component, the dynamic height applied by the player/builder is overridden by the default. 

In the screenshot below, the iframe element has the `height` attribute set to 136 (136px) but this style is overridden by the higher specificity styling given by the styled MUI wrapper (`.css-...` class).

Screenshot 2024-06-03 at 11 11 05

### How to fix

We should not apply a specific `height` when the element supports dynamic resizing.

spaenleh:
fix #885


ReidyT (approved):
LGTM 🚀

pyphilia (approved):

PR #904, opened 2024-06-17
fix: add timestamp to app url

### LINKED ISSUE #903: index.html can be aggresively cached by the browser causing issues
When the builder or player request an app, they set an iframe with the `src` set to the url of the `index.html` file of the app. A query param of `itemId` is added to let the app know what itemId is linked. 

We have observed that some browsers (Arc notably) have aggressive caching strategies where the index.html file is retrieved from disk cache. This is an issue when updating the app, as the index.html file references the JS files. Since the JS files have changed, the hash they wear has changed and if the browser uses an out of date version of the html file, it will request JS files that no longer exist. 

In order to fix this issue client side we can add a query parameter that will change frequently enough to counteract the aggressive caching strategy of the browser. 

Simply adding a query param that holds the current timestamp should fix this issue.

spaenleh:
fix #903

### DIFF
--- a/src/items/AppItem.tsx
+++ b/src/items/AppItem.tsx
@@ -88,12 +88,18 @@ const AppItem = ({
   const onLoad = (): void => setIsIFrameLoading(false);
 
   const appUrlWithQuery = useMemo(
-    () => appendQueryParamToUrl(appUrl, { itemId: item.id }),
+    () =>
+      appendQueryParamToUrl(appUrl, {
+        itemId: item.id,
+        // this ensures that the index.html can not be aggressively cached by the browser
+        ts: Date.now().toString(),
+      }),
     [item],
   );
 
   const iframe = (
     <AppIFrame
+      data-testid={frameId}
       id={frameId}
       ref={iFrameRef}
       isResizable={isResizable}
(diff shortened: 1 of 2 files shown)


src/items/AppItem.tsx
```diff
@@ -88,12 +88,18 @@ const AppItem = ({
   const onLoad = (): void => setIsIFrameLoading(false);
 
   const appUrlWithQuery = useMemo(
-    () => appendQueryParamToUrl(appUrl, { itemId: item.id }),
+    () =>
+      appendQueryParamToUrl(appUrl, {
+        itemId: item.id,
+        // this ensures that the index.html can not be aggressively cached by the browser
+        ts: Date.now().toString(),
```
ReidyT, line 95: Maybe `ts` is not explaining very well the purpose. What do you think about something like:

```suggestion
        timestampVersion: Date.now().toString(),
```

ReidyT (approved):
LGTM 😶‍🌫️ I just let a comment about the name of the new param.

src/items/AppItem.tsx
spaenleh: @ReidyT Thank you for the comment, yes the param name is not very cleat, however its name does not really matter, only its presence and the fact that it changes often enough for the browser to not be able to cache the request is actually useful. See [this article](https://stackoverflow.com/questions/67219450/why-does-adding-a-random-get-parameter-make-favicon-show) on favicon. So maybe I could use `v` instead ? I think making it short is important too.

PR #911, opened 2024-06-21
feat: add file uploader

pyphilia:
Screenshot 2024-06-21 at 18 01 39
Screenshot 2024-06-21 at 18 01 48


src/FileDropper/FileDropper.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+      },
+      action: 'on drop',
+    },
+  },
+} satisfies Meta<typeof FileDropper>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
```
spaenleh, line 34: You can also use the `satisfies Story` syntax for exported stories.

src/FileDropper/FileDropper.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+  args: {
+    onChange: () => {
+      console.log('on change');
+    },
+    onDrop: () => {
+      console.log('woiejfkl');
+    },
+  },
```
spaenleh, line 42: You can set default arguments in the meta, for functions use `fn()` that you import from `import { fn } from '@storybook/test';`. This is useful so tests do not complain.

src/FileDropper/FileDropper.tsx
```diff
@@ -0,0 +1,193 @@
...
+  Typography,
+  useTheme,
+} from '@mui/material';
+
+import { DndProvider, DropTargetMonitor, useDrop } from 'react-dnd';
+import { HTML5Backend, NativeTypes } from 'react-dnd-html5-backend';
+
+import UploadFileButton, {
+  UploadFileButtonProps,
+} from '../buttons/UploadFileButton/UploadFileButton';
```
spaenleh, line 18: I understand why you put the Upload button under the `buttons` folder, but I do not really think that `button` is a "feature". I would prefer for the buttons to be colocated with their feature. Not something we need to do now, but here I think it would make more sens to have the upload button in the same folder.

src/FileDropper/FileDropper.tsx
```diff
@@ -0,0 +1,193 @@
...
+   */
+  releaseText?: string;
+  /**
+   * Whether files are getting uploaded
+   */
+  isLoading?: boolean;
+  /**
+   * If `isLoading` is true, shows given `upload progress` value in a progress bar
+   */
+  uploadProgress?: number;
```
spaenleh, line 66: Would it be possible to combine these two props into a single one ? 
If using `uploadProgress` only you could derive a local variable `isLoading` that would be true when uploadProgress is not undefined.

src/FileDropper/FileDropper.tsx
```diff
@@ -0,0 +1,193 @@
...
+      alignItems='center'
+      bgcolor={error ? '#ffbaba' : bgColor}
+      borderRadius={5}
+      ref={drop}
+      gap={2}
+      sx={{
+        borderStyle: 'dashed',
+        borderWidth: 3,
+        borderColor: error ? 'red' : 'lightgrey',
+        ...sx,
```
spaenleh, line 125: Is it really necessary to explose a `sx` ? It feels like we just do not know how it should be re-used, so we stuff a "put everything you want" prop to account for that.

src/FileDropper/FileDropper.tsx
```diff
@@ -0,0 +1,193 @@
...
+    bgColor = '#ffbaba';
+  }
+
+  return (
+    <Stack
+      role='dropzone'
+      id={id}
+      height='100%'
+      width='100%'
+      display='flex'
```
spaenleh, line 114: No need for `display="flex"` on `Stack` it is already set by default.

src/FileDropper/FileDropper.tsx
```diff
@@ -0,0 +1,193 @@
...
+      sx={{
+        borderStyle: 'dashed',
+        borderWidth: 3,
+        borderColor: error ? 'red' : 'lightgrey',
+        ...sx,
+      }}
+      py={4}
+      boxSizing='border-box'
+    >
+      <Stack display='block'>
```
spaenleh, line 130: Why ?? Use a `Box` no? is it really necessary ?

src/FileDropper/FileDropper.tsx
```diff
@@ -0,0 +1,193 @@
...
+          </Stack>
+          {hints && <Typography variant='caption'>{hints}</Typography>}
+          {error && <Alert severity='error'>{error}</Alert>}
+        </>
+      )}
+    </Stack>
+  );
+};
+
+const FileDropper = (args: FileDropperProps): JSX.Element | null => {
```
spaenleh, line 170:
```suggestion
const FileDropper = (args: FileDropperProps): JSX.Element => {
```

src/buttons/UploadFileButton/UploadFileButton.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+      action: 'on change',
+    },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof UploadFileButton>;
+
+export const Default: Story = {
```
spaenleh, line 26: Use `satisfies Story`.

src/buttons/UploadFileButton/UploadFileButton.tsx
```diff
@@ -0,0 +1,97 @@
+import { CloudUploadIcon } from 'lucide-react';
+
+import { CircularProgress, styled } from '@mui/material';
+
+import { ChangeEventHandler } from 'react';
+
+import GraaspButton, { GraaspButtonProps } from '../Button/Button';
+
+export interface UploadFileButtonProps {
```
spaenleh, line 9: Use `type`.

src/buttons/UploadFileButton/UploadFileButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+  id?: string;
+  size?: GraaspButtonProps['size'];
+  /**
+   * button color
+   */
+  color?: GraaspButtonProps['color'];
+  /**
+   * button variant
+   */
+  variant?: GraaspButtonProps['variant'];
```
spaenleh, line 19: Would it make sens to write it as 
```tsx
type Props = {
  id?: string
  onChange: ...
} & GraaspButtonProps

```

src/buttons/UploadFileButton/UploadFileButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+  icon = <CloudUploadIcon />,
+  size,
+  color,
+}: UploadFileButtonProps): JSX.Element => (
+  <GraaspButton
+    id={id}
+    component='label'
+    variant={variant}
+    color={color}
+    startIcon={
```
spaenleh, line 81: There is the `LoadingButton` that you can get from `@mui/lab` that does the loading indicator automatically.

spaenleh (changes requested):
Looks good, I have added some comments related to style and slightly modifying the props. 

Really like the work you are doing with these buttons to simplify and make them available  elsewhere.

src/FileDropper/FileDropper.tsx
pyphilia: Ok, so I created a `upload` folder where I've put both FileDropper and UploadButton.

spaenleh (approved):
Looks good, I have added some nitpicks, I don't want to bother on form too much, let me know if you want to see how we can improve. Otherwise it is fine to merge like this.

PR #914, opened 2024-06-25
feat: add draggable elements, to replace table

pyphilia:
This PR adds `DraggingWrapper` to allow moving and dropping elements in other elements. I changed the cards for updated item table layout.

### DIFF
--- a/src/Card/Card.tsx
+++ b/src/Card/Card.tsx
@@ -1,92 +1,196 @@
-import { Box, Stack, SxProps, styled } from '@mui/material';
-import Card from '@mui/material/Card';
+import { Stack, SxProps, styled } from '@mui/material';
+import MuiCard from '@mui/material/Card';
 import CardActions from '@mui/material/CardActions';
 import Typography from '@mui/material/Typography';
+import Grid2 from '@mui/material/Unstable_Grid2';
 
-import { FC, ReactElement } from 'react';
+import { ReactElement } from 'react';
+import { Link } from 'react-router-dom';
 
-import CardHeader from './CardHeader';
+import type { DraggableAndDroppableProps } from '../draggable/types';
+import CardThumbnail, { CardThumbnailProps } from './CardThumbnail';
 
 const DEFAULT_CARD_HEIGHT = 130;
 
-const StyledImage = styled('img')({
-  width: '100%',
-  height: '100%',
-  objectFit: 'cover',
-});
-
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -77,6 +77,8 @@
     "@ag-grid-community/client-side-row-model": "31.3.2",
     "@ag-grid-community/react": "^31.3.2",
     "@ag-grid-community/styles": "^31.3.2",
+    "@faker-js/faker": "8.3.1",
+    "@graasp/sdk": "4.14.1",
```
spaenleh, line 81: If you need these when developing, you should add them to the `devDependencies` I am not sure they need to be direct dependencies of the package.

src/Card/Card.tsx
```diff
@@ -1,92 +1,202 @@
...
+}: {
+  children?: string | JSX.Element | JSX.Element[];
+  to?: string;
+}): JSX.Element => {
+  if (!to) {
+    return <>{children}</>;
+  }
+
+  return (
+    <Link to={to} style={{ textDecoration: 'none', color: 'unset' }}>
```
spaenleh, line 70: Use `styled` ?

src/draggable/DraggingWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+  renderComponent: DraggableElementProps<T>['renderComponent'];
+
+  rows?: T[];
+
+  /** wrapper id */
+  id?: string;
+  getRowId?: (el: T) => string;
+
+  /** show checkbox */
+  // showCheckbox?: boolean;
```
spaenleh, line 23: Remove ?

src/draggable/DraggingWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+  rows = [],
+  getRowId,
+  onDropInRow: onDropInRowFn,
+  onDropBetweenRow: onDropBetweenRowFn,
+  renderComponent,
+  isMovable,
+  enableMoveInBetween = true,
+  allowFiles = true,
+  canDrop,
+  xs = 12,
```
spaenleh, line 51: How are you planning to use this property ? If it is to create more columns depending on the breakpoint, directly set the number based on the breakpoint.

spaenleh (approved):
Looks good! I still have to see it live, as it is a bit hard to represent only with the code.

PR #959, opened 2024-07-17
feat: add card selection props

pyphilia:


src/Card/Card.tsx
```diff
@@ -14,20 +16,28 @@ const DEFAULT_CARD_HEIGHT = 130;
...
+    : isSelected
+      ? `2px solid ${PRIMARY_COLOR}`
+      : 'none',
+}));
 
 type CardProps = {
   name: string | JSX.Element;
   alt: string;
   id?: string;
+  className?: string;
```
spaenleh, line 40: Why the use of `className` ?

src/Card/Card.tsx
```diff
@@ -14,20 +16,28 @@ const DEFAULT_CARD_HEIGHT = 130;
 
 const StyledCard = styled(MuiCard, {
   shouldForwardProp: (prop) => prop !== 'elevation' && prop !== 'fullWidth',
```
spaenleh, line 18: I think you need to add some props here to remove warnings that they are not supported on the DOM. `isSelected` and `isOver` for example

spaenleh (changes requested):

src/Card/Card.tsx
pyphilia: I needed a generic way to select all cards, not based on the id https://github.com/graasp/graasp-builder/blob/f7cc3a22b17a8bbf4b587ebd93b1eb34243cc221/src/components/main/list/SelectionContext.tsx#L95

spaenleh (approved):

src/Card/Card.tsx
```diff
@@ -13,21 +15,26 @@ import CardThumbnail, { CardThumbnailProps } from './CardThumbnail';
...
-  ({ theme, elevation, fullWidth, isOver }) => ({
-    borderRadius: theme.spacing(1),
-    boxShadow: elevation ? theme.shadows[2] : '0px 2px 2px #eeeeee',
-    width: fullWidth ? '100%' : 'max-content',
-    maxWidth: '100%',
-    border: isOver ? '2px solid black' : 'none',
-  }),
-);
+  shouldForwardProp: (prop: string) =>
+    !['elevation', 'fullWidth', 'isSelected', 'isOver'].includes(prop),
```
ReidyT, line 19: Should we extract the array in a const ?

```suggestion
    !PROPS_TO_FORWARD.includes(prop),
```

ReidyT (approved):
Nice PR, LGTM!

src/Card/Card.tsx
```diff
@@ -13,21 +15,26 @@ import CardThumbnail, { CardThumbnailProps } from './CardThumbnail';
...
+  isOver: boolean;
+  fullWidth?: boolean;
+  elevation?: boolean;
+  isSelected?: boolean;
+}>(({ theme, elevation, fullWidth, isOver, isSelected }) => ({
+  borderRadius: theme.spacing(1),
+  boxShadow: elevation ? theme.shadows[2] : '0px 2px 2px #eeeeee',
+  width: fullWidth ? '100%' : 'max-content',
+  maxWidth: '100%',
+  border: isOver || isSelected ? `2px solid ${PRIMARY_COLOR}` : 'none',
```
spaenleh, line 30: Using `outline` instead of `border` will be better as the element will not change size, removing small layout shifts.

spaenleh (changes requested):

spaenleh (approved):
LGTM ! 🎈

PR #970, opened 2024-07-30
feat: use ts build

spaenleh:
In this PR I convert the package to be built with Typescript only and not vite library mode.

### Motivation

While vite library mode is very convenient to create a bundled package, it uses single file mode by default which makes the library code unable to be tree-shaked. 

If we expose the library with preserved module structure, the library should be tree-shakable and the consumers should be able to reduce the size of their bundle by not including unused code. 

### Caveats

In my test, after exposing the Typescript build, the size of the consumer did not really decrease. 
Possible reasons for this is that since I had to use the named imports of @mui/material instead of the second level default exports The bundle was still larger. This issue seems to not be on the fix list for the MUI team, as seems like a design issue on their behalf. If we wanted to really reduce our bundle size we should look at another component library that has better support for bundle size optimisation and tree-shaking. If looking at the analysis from https://arethetypeswrong.github.io/?p=@mui/material@5.16.5 the issue lies in the fact that the second level imports are maskaraded in CJS exports. So it will not play nicely with the current ts + nodenext setup. 

### Other changes

I took the opportunity to do a bit of summer cleaning too:
- use as much `lucide-icons` as possible to replace the mui icons
- use `type` instead of `interface` in all prop definitions
- remove the usage of `FC` in favour of `(props: Props) => JSX.Element` 
- remove intermediate `index.ts` files in subfolders and directly export the components in the main `index.ts`

### Questions

Should we use only named exports for components instead of having default exports that are then named ?

#### Player: 
https://github.com/graasp/graasp-player/pull/825 
2,312.19 kB -> 1,589.27 kB
31% shrink in bundle size

#### Builder
https://github.com/graasp/graasp-builder/pull/1390
4,646.32 kB -> 4,379.26 kB
5% shrink in bundle size

#### Library
https://github.com/graasp/graasp-library/pull/665
1.56Mb -> 1.45 Mb
7% shrink in bundle size

#### Analytics
https://github.com/graasp/graasp-analytics/pull/419
2,257.06 kB -> 2,033.90 kB
9% shrink in bundle size

#### Account 
https://github.com/graasp/graasp-account/pull/315
1,660.81 kB -> 1,351.62 kB
18% shrink in bundle size

#### Auth
https://github.com/graasp/graasp-auth/pull/442
1,599.91 kB -> 1,280.01 kB
19% shrink in bundle size


ReidyT (approved):
Thanks for the PR, LGTM! 👾

src/CreativeCommons/icons/CCIconsProps.js
```diff
@@ -0,0 +1 @@
+export {};
```
pyphilia, line 1: Necessary?

src/itemLogin/utils.ts
```diff
@@ -1,30 +1,5 @@
...
-    if (els.length <= 1) {
-      return [];
-    }
-    [p] = els;
-  }
-  const ids = p.replace(/_/g, '-').split('.');
-  return ids;
-};
-
 export const isMemberIdValid = (memberId?: string): boolean => {
```
pyphilia, line 3: Is this actually used?

pyphilia (approved):

src/CreativeCommons/icons/CCIconsProps.js
spaenleh: This was an artefact from a failed build. I removed it.

PR #1003, opened 2024-08-28
fix: use graasp version of stylis to be compatible with esm

spaenleh:
In this PR I fix a bug introduced in #970 that made the interface crash when the language was set to arabic.

This was because of an issue with commonJS packages providing a default export for compatibility with ESM that actually does not work in ESM projects using NodeNext module resolution. 

The offending line was: https://github.com/graasp/graasp-ui/blob/18ac53187cbe4b6ce35a4c1744074a4d5e1efefe/src/ThemeContext/ThemeContext.tsx#L45

To resolve the issue I forked the `stylis-plugin-rtl` (no updates since 3 years) and updated the build to provide a correct implementation of the types and modules. This was verified by using the excellent https://arethetypeswrong.github.io/ tool. Below are the results of the analysis for the original package and for my fork.

Analysis of the original package: https://arethetypeswrong.github.io/?p=stylis-plugin-rtl@2.1.1
Screenshot 2024-08-28 at 08 10 35

Analysis of my fork: https://arethetypeswrong.github.io/?p=@graasp/stylis-plugin-rtl@2.2.0
Screenshot 2024-08-28 at 08 15 22


ReidyT (approved):
Great job, really interesting 🔥

pyphilia (approved):

PR #1005, opened 2024-08-28
feat: add new funtion to translate a username into a unique color

spaenleh:


src/Avatar/stringToColor.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Generate a background color for avatars from the given string
```
pyphilia, line 2: Is there a limitation as being a "background"? like it's always readable if you set a black text on it? Just wondering. If not maybe you can generalise the comment.

pyphilia (approved):
Great! Just a suggestion: if it's not so tight to avatars, I would move the files in utils.

ReidyT (approved):
LGTM 👍

PR #1016, opened 2024-09-11
fix: improve downloadButton

### LINKED ISSUE #920: Download button for file displays the original file name instead of the item name
# Bug 🐛 

**Renommer un fichier dans le builder ne change pas le nom dans le player**

## Expected Behavior
Changer le nom d'un fichier dans le builder devrait changer le nom du fichier sur le bouton "download" dans le player

## Current Behavior
Je change le nom d'un fichier via le builder. Lorsque je vais sur la player, le nom du fichier n'a pas changé sur le bouton "download". Le nom est toujours celui que le fichier avait lors de l'importation.

## Steps to Reproduce
1) Importer un fichier dans le builder, par exemple un fichier excel (.xlsx)
2) Changer le nom du fichier grâce au bouton "modifier" ( le crayon)
3) Cliquer sur le bouton player
4) A l'intérieur du bouton "download", le nom du fichier n'a pas changé

## Context (Environment)
Mon système d'exploitation est: MacOS Big Sur version 11.6.2
Mon browser est: google chrome
J'utilise la version Graasp FR. Le bug est aussi sur la version EN

spaenleh:
Before:
Screenshot 2024-09-12 at 15 36 41

Now:
Screenshot 2024-09-12 at 15 23 16

fix #920

### DIFF
--- a/src/items/DownloadButtonFileItem.stories.tsx
+++ b/src/items/DownloadButtonFileItem.stories.tsx
@@ -1,0 +1,22 @@
+import { Meta, StoryObj } from '@storybook/react';
+
+import DownloadButtonFileItem from './DownloadButtonFileItem.js';
+
+const meta = {
+  title: 'Items/FileItemDownload',
+  component: DownloadButtonFileItem,
+  args: {
+    url: '#',
+  },
+} satisfies Meta<typeof DownloadButtonFileItem>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Button = {
+  args: {
+    name: 'My powerpoint presentation',
+    caption: '(50Mb - application/pdf)',
+  },
+} satisfies Story;
--- a/src/items/DownloadButtonFileItem.stories.tsx
+++ b/src/items/DownloadButtonFileItem.stories.tsx
@@ -17,6 +17,6 @@ type Story = StoryObj<typeof meta>;
 export const Button = {
   args: {
     name: 'My powerpoint presentation',
-    caption: '(50Mb - application/pdf)',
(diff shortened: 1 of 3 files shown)


ReidyT (approved):
Really nice, LGTM! I just have two questions:
1) It is possible to display the extension even with the item name?
2) It is possible to display a special icon depending on the file? (pdf icon, ppt icon, etc). I think this one can be in another issue.

spaenleh:
@ReidyT 
1. I was thinking about adding an indication for the format of the file (I can get it from the mimetype). But not as an extension, more like on the second line next to the size ? Or we could display the original file name, but there is no way for the user to change it, so it would not be a good design I think. 
2. Since for the moment we are showing this button for file formats we do not support, it would be difficult to display the right logo, but I can see a use where people want a supported element to be displayed as a downloadable element (audio already allows to donwload via the native element), but imagine you want to display the pdf as a button and not as an embed.

ReidyT:
> @ReidyT
> 
> 1. I was thinking about adding an indication for the format of the file (I can get it from the mimetype). But not as an extension, more like on the second line next to the size ? Or we could display the original file name, but there is no way for the user to change it, so it would not be a good design I think.
> 2. Since for the moment we are showing this button for file formats we do not support, it would be difficult to display the right logo, but I can see a use where people want a supported element to be displayed as a downloadable element (audio already allows to donwload via the native element), but imagine you want to display the pdf as a button and not as an embed.

1. Yeah I think it could be a good idea. On my point of view, it's always interesting to know the type of the file you'll download.
2. I understand that it's not easy, it was more a potential improvement 😄.

PR #1045, opened 2024-09-17
fix(app): use memberId for legacy apps

pyphilia:
UI applies the new namenclature `accountId`, and itself adds a property `memberId` so it is compatible with legacy apps.

### DIFF
--- a/src/items/appItemHooks.ts
+++ b/src/items/appItemHooks.ts
@@ -126,7 +126,6 @@ const useAppCommunication = ({
 
             payload: {
               /**
-               * @deprecated use accountId.
                * Legacy for old apps or apps that does not use apps-query-client
                */
               memberId: contextPayload.accountId,


ReidyT (approved):
LGTM 👍

src/items/appItemHooks.ts
```diff
@@ -129,7 +123,15 @@ const useAppCommunication = ({
         iFrameRef?.current?.contentWindow?.postMessage(
           JSON.stringify({
             type: POST_MESSAGE_KEYS.GET_CONTEXT_SUCCESS,
-            payload: contextPayload,
+
+            payload: {
+              /**
+               * @deprecated use accountId.
```
spaenleh, line 129: Not sure we need to mark this as deprecated ? But I think it is important to keep the comment that explains why this is here.

PR #1046, opened 2024-09-17
feat: refactor with authorization, change current account type

pyphilia:
In this PR:
- withAuthorization -> SignedInWrapper (Component)
- ItemLoginWrapper (Component)
- PreventGuestWrapper (Component) - new
- update `currentMember` type

We can discuss the naming.


src/itemLogin/ItemLoginAuthorization.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const LogInForm: Story = {
+  args: {
+    itemLoginSchemaType: ItemLoginSchemaType.Username,
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+
+    expect(canvas.getByText('Sign In')).toBeVisible();
+  },
+};
```
spaenleh, line 52: This is ok, but I prefer to use the following style: 

```suggestion
export const LogInForm = {
  args: {
    itemLoginSchemaType: ItemLoginSchemaType.Username,
  },
  play: async ({ canvasElement }) => {
    const canvas = within(canvasElement);

    expect(canvas.getByText('Sign In')).toBeVisible();
  },
} satisfies Story;
```

spaenleh (approved):
Looks good ! I have left a few comments.

src/Authorization/PreventGuestWrapper.tsx
```diff
@@ -0,0 +1,73 @@
...
+  currentAccount,
+  error,
+  id,
+  onButtonClick,
+  startIcon = <ClipboardPen />,
+  errorText = 'An error occured.',
+  text = 'You are currently using Graasp with a guest account. In order to use all features of Graasp, you have to log out and create a Graasp account.',
+}: Props): ReactNode => {
+  if (currentAccount) {
+    // guest - should not have access to home
```
ReidyT, line 44:
```suggestion
    // guest - should not have access to children component
```

src/Authorization/PreventGuestWrapper.tsx
```diff
@@ -0,0 +1,73 @@
...
+  error?: JSX.Element;
+  errorText?: string;
+  id?: string;
+  onButtonClick?: () => void;
+  startIcon?: JSX.Element;
+  text?: string | JSX.Element;
+};
+
+const PreventGuestWrapper = ({
+  buttonText = 'Log out and Create a Graasp account',
```
ReidyT, line 33: I'm wondering if it's necessary to have defaults and to not define as required parameters. In any case, we'll have to redefine their to use the correct language right? Having their as optional add a risk to forgetting to define their with the user's language.

src/Authorization/SignedInWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const SignedInWrapper = ({
+  currentAccount,
+  redirectionLink,
+  onRedirect,
+  children,
+}: SignedInWrapperProps): SignedInWrapperProps['children'] => {
+  const redirectToSignIn = (): void => {
+    if (!redirectionLink) {
+      return console.debug('No link has been set for redirection');
```
ReidyT, line 22:
```suggestion
     console.debug('No link has been set for redirection');
     return;
```

src/Authorization/PreventGuestWrapper.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </div>
+    ),
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof PreventGuestWrapper>;
+
+export const Guest: Story = {
```
ReidyT, line 30: Should we also use `satisfies Story` here to be consistent?

src/itemLogin/ItemLoginAuthorization.tsx
```diff
@@ -19,99 +14,58 @@ import ItemLoginScreen, { SignInPropertiesType } from './ItemLoginScreen.js';
...
+  }
 
-      // the item could be fetched without errors
-      // because the user is signed in and has access
-      // or because the item is public
-      if (item && item.id) {
-        return <ChildComponent />;
-      }
+  // signed out but can sign in with item login
+  if ((!currentAccount || !currentAccount.id) && itemLoginSchemaType) {
```
ReidyT, line 53: This works too right?

```suggestion
  if (!currentAccount?.id) && itemLoginSchemaType) {
```

ReidyT (dismissed):
LGTM 🧑‍🚀 I wonder if we should make the optional parameters of `PreventGuestWrapper` mandatory. What do you think (I elaborate that directly in the code review)?

PR #1050, opened 2024-09-18
fix: quill placeholder does not update

### LINKED ISSUE #1049: [BUG] React-Quill `placeholder` props does not update (is not translated eiter)
## Description

The translated placeholder text is not applied to the textarea.

## Explanation of the bug

According to [an old issue in react-quill](https://github.com/zenoamaro/react-quill/issues/340) a change of the `placeholder` prop does not cause its update in the quill editor.

The solution that is exposed by the authors in 2018 was to use a ref and a useEffect. Why they did not consider adding this in the react component themselves stays a mystery to me...

spaenleh:
This PR fixes a very sneaky bug with the unusual behaviour of the `placeholder` prop from react-quill.

I have added an interaction test that ensures that the placeholder can be updated.

fix #1049

### DIFF
--- a/src/TextEditor/TextEditor.stories.tsx
+++ b/src/TextEditor/TextEditor.stories.tsx
@@ -18,7 +18,6 @@ const meta = {
   args: {
     onChange: fn(),
   },
-  render: (args) => <TextEditor {...args} />,
 } satisfies Meta<typeof TextEditor>;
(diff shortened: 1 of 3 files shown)


pyphilia (approved):
LGTM, looks like a fun bug 💀

PR #1062, opened 2024-09-24
fix: pass classname to dense card

pyphilia:
ref https://github.com/graasp/graasp-builder/issues/1433

The solution is to keep track of cards, which is done through a class for dense cards, but not non-dense cards.


ReidyT (approved):
Thanks for this fix, LGTM 🌞 Maybe just add a comment to explain why we pass the class?

PR #1067, opened 2024-09-30
fix: update ThemeProvider children type to be used with conditional rendering

spaenleh:
The current type does not allow to have `null` elements as children. If we want to be able to use condition rendering inside the children object we should use the more loose type of `ReactNode` to type the children prop.


ReidyT (approved):
LGTM 🦤

PR #1076, opened 2024-10-03
fix: children type and umami-tags

spaenleh:
In this PR:
- update the type of the children prop to `ReactNode` to allow all possible valid react elements to be a children
- add some umami tags for use in builder


ReidyT (approved):
LGTM 🍟

PR #1080, opened 2024-10-07
feat: add umami for main components

spaenleh:


src/MainMenu/MenuItem/MenuItem.tsx
```diff
@@ -46,6 +46,7 @@ export const MenuItem = ({
         onClick={onNavigate}
         disabled={disabled}
         selected={selected}
+        data-umami-event={id}
```
pyphilia, line 49: I think here I would prefer to get `umamiEvent` from the props. Sidebar decides what should be the event shape.

src/UserSwitch/UserSwitch.tsx
```diff
@@ -194,6 +194,7 @@ export const UserSwitch = ({
         aria-controls={menuId}
         aria-expanded={Boolean(anchorEl)}
         tabIndex={0}
+        data-umami-event='header-avatar'
```
pyphilia, line 197: Should this be hardcoded? the userswitch could be used somewhere else than the header.

src/PlatformSwitch/PlatformSwitch.tsx
```diff
@@ -134,6 +134,7 @@ export const PlatformSwitch = ({
           data-testid={platform}
           href={(!platformProps?.disabled && platformProps?.href) || '#'}
           aria-disabled={platformProps?.disabled}
+          data-umami-event={`header-navigation-switch-${platform}`}
```
pyphilia, line 137: Could be use somewhere else than the header?

src/SearchInput/SearchInput.tsx
```diff
@@ -42,7 +51,12 @@ export const SearchInput = ({
       sx={{ flex: 1, width: width ?? '100%', backgroundColor: 'white' }}
       placeholder={placeholder}
       size={size}
-      inputProps={{ 'aria-label': ariaLabel }}
+      inputProps={{
+        'aria-label': ariaLabel,
+        // Umami data props
+        'data-umami-event': dataUmami?.event,
+        'data-umami-event-page': dataUmami?.page,
```
pyphilia, line 58:
```suggestion
        // Umami data props
        {...buildUmamiProps(dataUmami)}
```
Not fan of spread props tho, but you're then free to add context if you need later.

pyphilia (changes requested):
LGTM, maybe check comments on builder first.

src/UserSwitch/UserSwitch.tsx
```diff
@@ -194,6 +196,7 @@ export const UserSwitch = ({
         aria-controls={menuId}
         aria-expanded={Boolean(anchorEl)}
         tabIndex={0}
+        data-umami-event={dataUmami}
```
pyphilia, line 199:
```suggestion
        data-umami-event={dataUmamiEvent}
```

pyphilia (approved):
Thanks, same as in Builder!

PR #1091, opened 2024-10-17
fix: footer should use footer html tag

spaenleh:
This fixes a small bug where the footer of the `Main` component was using a `main` html tag instead of a `footer` tag.


ReidyT (approved):
Nice!

PR #1121, opened 2024-11-25
fix: remove testing library addon and use unified testing package for storybook

spaenleh:
In this PR: 
- remove `@storybook/testing-library` package
- use unified `@storybook/test` package instead


ReidyT (approved):
🤩

